PR #33, opened 2019-08-29
Support PATH native-image on *nix-likes

sogaiu:
Attempt to use native-image from PATH on *nix-like systems.

Tested locally on machine with native-image on PATH. I don't have a local setup to fully test the other path -- where there is no native-image on PATH.

### DIFF
--- a/script/compile
+++ b/script/compile
@@ -2,12 +2,18 @@
 
 set -eo pipefail
 
-if [ -z "$GRAALVM_HOME" ]; then
-    echo "Please set GRAALVM_HOME"
-    exit 1
-fi
+NATIVE_IMAGE=`which native-image` || true
 
-"$GRAALVM_HOME/bin/gu" install native-image || true
+if [ -z "$NATIVE_IMAGE" ]; then
+  if [ -z "$GRAALVM_HOME" ]; then
+      echo "Please set GRAALVM_HOME"
+      exit 1
+  fi
+
+  "$GRAALVM_HOME/bin/gu" install native-image || true
+
+  NATIVE_IMAGE="$GRAALVM_HOME/bin/native-image"
+fi
 
 BABASHKA_VERSION=$(cat resources/BABASHKA_VERSION)
 
@@ -18,7 +24,7 @@ BABASHKA_VERSION=$(cat resources/BABASHKA_VERSION)
 # cp -R /tmp/sci/src/* src
 
 lein with-profiles +clojure-1.10.1 do clean, uberjar
-$GRAALVM_HOME/bin/native-image \
+$NATIVE_IMAGE \
   -jar target/babashka-$BABASHKA_VERSION-standalone.jar \
   -H:Name=bb \
   -H:+ReportExceptionStackTraces \


borkdude:
@sogaiu Did you perhaps also run the unit tests using the native binary?

```
BABASHKA_TEST_ENV=native BABASHKA_PRELOADS='(defn __bb__foo [] "foo") (defn __bb__bar [] "bar")' lein test
```

I think the SSL test may fail if you don't set `GRAALVM_HOME`

```
(deftest ssl-test
  (let [graalvm-home (System/getenv "GRAALVM_HOME")
        lib-path (format "%1$s/jre/lib:%1$s/jre/lib/amd64" graalvm-home)
        ;; _ (prn "lib-path" lib-path)
        resp (bb nil (format "(System/setProperty \"java.library.path\" \"%s\")
                              (slurp \"https://www.google.com\")"
                             lib-path))]
    (is (re-find #"doctype html" resp))))
```

sogaiu:
No, I didn't :sweat_smile: 

ATM, this is what I get for `sh script/test` on master:

```
sh script/test
==== Testing JVM version
==== Testing JVM version

lein test babashka.impl.socket-repl-test
Babashka socket REPL started at 0.0.0.0:1666

lein test :only babashka.impl.socket-repl-test/socket-repl-test

FAIL in (socket-repl-test) (socket_repl_test.clj:40)
ctrl-d exits normally, doesn't print nil
expected: (str/ends-with? (:out (sh "bash" "-c" (if mac? "echo \"(inc 1336)\" | nc 127.0.0.1 1666" "echo \"(inc 1336)\" | nc -q 1 127.0.0.1 1666"))) "1337\nbb=> ")
  actual: (not (str/ends-with? "" "1337\nbb=> "))

lein test babashka.main-test
```

...and then hanging.

borkdude:
Might be worth looking into, especially regarding this issue:

```
lein test :only babashka.main-test/ssl-test
```

sogaiu:
When I tried:

```
GRAALVM_HOME=/usr/lib/jvm/java-8-graal \
BABASHKA_TEST_ENV=native \
BABASHKA_PRELOADS='(defn __bb__foo [] "foo") (defn __bb__bar [] "bar")' \
lein test
```

I got:

```
==== Testing native version
==== Testing native version

lein test babashka.impl.socket-repl-test

lein test :only babashka.impl.socket-repl-test/socket-repl-test

FAIL in (socket-repl-test) (socket_repl_test.clj:40)
ctrl-d exits normally, doesn't print nil
expected: (str/ends-with? (:out (sh "bash" "-c" (if mac? "echo \"(inc 1336)\" | nc 127.0.0.1 1666" "echo \"(inc 1336)\" | nc -q 1 127.0.0.1 1666"))) "1337\nbb=> ")
  actual: (not (str/ends-with? "" "1337\nbb=> "))

lein test babashka.main-test
```

and then hanging...which looks pretty similar to the jvm results.

Will try with just babashka.main-test/ssl-test next.

sogaiu:
For native:

```
$ GRAALVM_HOME=/usr/lib/jvm/java-8-graal BABASHKA_TEST_ENV=native BABASHKA_PRELOADS='(defn __bb__foo [] "foo") (defn __bb__bar [] "bar")' lein test :only babashka.main-test/ssl-test
==== Testing native version

lein test babashka.main-test

Ran 1 tests containing 1 assertions.
0 failures, 0 errors.
```

For jvm:

```
$ BABASHKA_PRELOADS='(defn __bb__foo [] "foo") (defn __bb__bar [] "bar")' lein test :only babashka.main-test/ssl-test

==== Testing JVM version

lein test babashka.main-test

Ran 1 tests containing 1 assertions.
0 failures, 0 errors.
```

So may be the ssl stuff is ok?

sogaiu:
It looks like the wait-for-it-test is hanging:

```
$ lein test :only babashka.main-test/wait-for-it-test
==== Testing JVM version

lein test babashka.main-test
```

https://github.com/borkdude/babashka/blob/4aa3021b14608d9218d23ff361230164fa513de6/test/babashka/main_test.clj#L171-L178

sogaiu:
Ah, of course...it's because Arch Linux and derivatives have the python binary pointed at python3.

When I change the test to say `python2`, it passes.

borkdude:
@sogaiu My point is that you changed the build script for not having to set `GRAALVM_HOME` when the binary is still on your path. But that approach doesn't work with the unit test for SSL yet.

sogaiu:
@borkdude Yes, I understand.  I don't yet have a solution :)  On a related note, even if you don't have the native-image binary on your path, the test is not necessarily going to work if you don't have GRAAL_VM set appropriately, right?

At least the hanging that was occuring appears to be explained now.

PR #60, opened 2019-09-06
Add an -e/--eval option, and automatically detect file name args

### LINKED ISSUE #59: introduce -e for expressions and interpret argument as file when it doesn't start with ( or *

plexus:
When an argument is a file name (i.e. the file exists), then treat it as a file,
rather than as an expression.

Update dev docs.

Closes #59

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,27 @@ $ bb '(slurp "https://www.clojure.org")' | bb '(subs *in* 0 50)'
 "<!doctype html><html itemscope=\"\" itemtype=\"http:/"
 ```
 
-## Test
+## Developing Babashka
+
+To work on Babashka itself make sure Git submodules are checked out.
+
+``` shellsession
+$ git clone https://github.com/borkdude/babashka --recursive
+```
+
+To update later on:
+
+``` shellsession
+$ git submodule update --recursive
+```
+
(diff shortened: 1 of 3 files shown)


borkdude:
@plexus Looks good to me. One minor thing:

> If neither -e, -f, or --socket-repl are specified, then the first argument is treated as a file if it exists, or as an expression otherwise.

It's actually the last expression that is treated like a file or expression. Could you ensure if this still works with your PR and change the text?

plexus:
Looking at your code it is the first argument that does not have a flag in front of it.

```shellsession
$ ./bb '(println 123)' '(println :abc)'
123
```

borkdude:
@plexus The first argument after known command line args is treated as a file or expression, if you don't use `-f`, etc. Everything after that is considered a command line arg for the expression.

plexus:
Exactly, that's why I said "the first" :) I'll try to rephrase, gotta prep for my live stream now.

borkdude:
@plexus I think the original phrasing wasn't clear as well, so I guess it's not something you should have to worry about. I'll merge and then see if it can be improved. Thanks for your contribution!

plexus:
:ok_hand:

plexus:
:ok_hand:

PR #78, opened 2019-10-13
Adapt net/wait-for-it as wait/wait-for-(port|path)

sogaiu:
Here's my initial attempt at wait/wait-for-port and wait/wait-for-path as discussed on slack a bit back.


borkdude:
Great! Can you also update the README?

sogaiu:
How does the README look now?

borkdude:
Looks good. Can you add tests for default for both wait-for-port and wait-for-path? Also I'd like to see a test for the non-timeout case of wait-for-path if that is possible.

sogaiu:
I can work on those things.

I'm a bit on the cautious side regarding creating and deleting files.  Is there a way within babashka to create temp files appropriately?  I saw that in one of the tests there is java.io.File/createTempFile -- but iiuc that's not exposed to babashka.

borkdude:
@sogaiu I think `java.io.File.createTempFile()` is not implemented yet, but now might be a good time to add it:

https://github.com/borkdude/babashka/blob/master/src/babashka/impl/File.clj

Alternatively you can just create temp files in `/tmp`. Be sure to delete them first, so the file won't be there from the previous test run.

sogaiu:
Thanks for the comments.

I'll investigate the createTempFile angle as I'm wary of messing up someone's data accidentally.

sogaiu:
Added createTempFile with a test and hopefully improved the wait-for-path testing.

sogaiu:
May be it would be better for createTempFile to exist as File/createTempFile since getenv is available as System/getenv.

PR #106, opened 2019-11-14
Added more String parsers

jeroenvandijk:
Added:

* Double/parseDouble
* Boolean/parseBoolean

I need this for a project where I parse command-line options with these types. I figured `Float/parseFloat` is redundant because of the implicit cast of Float to Double.

Note that I wasn't able to run the tests:

```
script/test
Syntax error (FileNotFoundException) compiling at (babashka/impl/socket_repl.clj:1:1).
Could not locate sci/impl/interpreter__init.class, sci/impl/interpreter.clj or sci/impl/interpreter.cljc on classpath.

Full report at:
/var/folders/ck/9zqvs0zx5vsf2bc7rvv1pwc40000gn/T/clojure-2853884920458488806.edn
Tests failed.
```
I've basically copied the `Integer.clj` namespace and adapted it to my best knowledge.

### DIFF
--- a/src/babashka/impl/Boolean.clj
+++ b/src/babashka/impl/Boolean.clj
@@ -1,0 +1,15 @@
+(ns babashka.impl.Boolean
+  {:no-doc true}
+  (:refer-clojure :exclude [list]))
+
+(set! *warn-on-reflection* true)
+
+(defn parseBoolean [^String x]
+  (Boolean/parseBoolean x))
+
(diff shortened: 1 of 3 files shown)


borkdude:
@jeroenvandijk To include sci, you need to also get the submodule, see:

https://github.com/borkdude/babashka#developing-babashka

PR looks ok to me!

Issue #103, opened 2019-11-14
Make it possible to write to a file via binding / .write / .append


holyjak:
Currently I do not see a way to write into a file in bb. In Clojure I'd do one of these

```clojure
(with-open [wrt (io/writer "my/file.txt")]
   (binding [*out* wrt] (println "hi"))
  (.write wrt "hi\n")
  (.append wrt "hi\n")
```

I am trying to change a number of files so writing to bb's stdout and redirecting to file won't help.

however none works in babashka. Or is there another way? Thank you!

borkdude:
What about `(spit file contents :append true)`?

PR #108, opened 2019-11-14
Add Writer.write, .append etc

holyjak:
Fixes part of #103


borkdude:
@holyjak Thanks!
Could you add a unit test for this? I'm refactoring Java interop at the moment and I want to make sure this isn't going to break your additions.

PR #141, opened 2019-12-07
Remove non-portable nc option

sogaiu:
An attempt to address https://github.com/borkdude/babashka/issues/140


sogaiu:
Might be better to not use netcat

PR #181, opened 2019-12-21
Expose the path to the running script as $0

j-cr:
Please feel free to discard, it's just for the sake of example.

- what the name should be?
- should a var be created, similar to `user/*input*`?


borkdude:
That works only when:
- You don't have a namespace form, i.e. stay within the user namespace
and doesn't work when you use `load-file` or `require`. As described in the issue, it requires more fundamental changes.

j-cr:
Makes sense. What I had in mind is shebang-style invokations only, but Joker's `*file*` behaves correctly in all the cases you described so it makes sense to mimic that.

PR #256, opened 2020-02-06
Noticed a typo while trying an example

cldwalker:
Thanks for the awesome library!


borkdude:
Thanks! Yes, that alias had been renamed.

PR #270, opened 2020-02-21
Adding java.time.temporal.ChronoUnit to classes.clj

mbjarland:
This enables the two following kinds of calculations between `LocalDateTime`, `LocalTime`, `Instant` and other `Temporal` descendants:

```
ChronoUnit.MILLIS.between(temporalOne, temporalTwo)
temporalOne.until(temporalTwo, Chronounit.HOURS)
```
(or corresponding clojure syntax) 

and I would assume a lot of other operations which require time units. Specifically I ran into a problem with calculating the number of millis between two `LocalDateTime` instances which this PR would solve.

### DIFF
--- a/src/babashka/impl/classes.clj
+++ b/src/babashka/impl/classes.clj
@@ -83,6 +83,7 @@
           java.time.ZonedDateTime
           java.time.ZoneId
           java.time.ZoneOffset
+          java.time.temporal.ChronoUnit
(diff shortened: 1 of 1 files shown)


borkdude:
@mbjarland Thanks a lot. Is is possible to encode this into a unit test?

borkdude:
Merged together with unit test via other PR.

PR #305, opened 2020-03-21
Support AWS Lambda custom runtime

dainiusjocas:
AWS Lambda doesn't support signal [handling](https://github.com/oracle/graal/issues/841), e.g. [here](https://github.com/quarkusio/quarkus/issues/4262). Therefore, in order for babashka to work, it is needed to disable the handling of SIGPIPE.

I propose to introduce an environment variable to alter the default behaviour and disable the handling of SIGPIPE: `BABASHKA_DISABLE_PIPE_HANDLER`. 

An example project that runs babashka script in AWS Lambda https://github.com/dainiusjocas/babashka-lambda.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1106,7 +1106,7 @@ Your command line args: (1 2 3)
 
 ## Package babashka script as a AWS Lambda
 
-AWS Lambda runtime doesn't support signals, therefore babashka has to disable handling of the SIGPIPE. This can be done by setting `DISABLE_PIPE_SIGNAL_HANDLING` to `true`.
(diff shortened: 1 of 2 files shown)


borkdude:
Without the change Lambda fails in a very similar way as  is described here https://github.com/oracle/graal/issues/841

borkdude:
So far babashka specific env variables starts with `BABASHKA_`,  like `BABASHKA_CLASSPATH`.
What about `BABASHKA_DISABLE_PIPE_HANDLER`?

PR #334, opened 2020-04-07
Added an example about using deps.clj script to produce the classpath

Rovanion:
I found the examples using aliases hard to understand and didn't want to spin up an entire clojure instance to calculate the classpath.

Feel free to do with it as you want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -455,6 +455,28 @@ $ bb --classpath src --main my.namespace
 Hello from my namespace!
 ```
 
+So if you have a larger script with a classic Clojure project layout like
+
+```shellsession
+$ tree -L 3
+├── deps.edn
+├── README
+├── src
+│   └── project_namespace
+│       ├── main.clj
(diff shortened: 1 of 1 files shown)


borkdude:
@Rovanion Isn't it much simpler to start bb like `bb -cp src:test --socket-repl 1666`?

Rovanion:
Yup, as long as there are no external dependencies that would be simpler. But I did not understand that doing so was a possibility from reading the README. Now that I see your example here I understand it though.

Perhaps that should be a section in the readme:

If you have a larger script with a classic Clojure structure like:

```
$ tree
...
```

Then you may start a socket repl like:

```
bb -cp src:test --socket-repl 1666
```

borkdude:
Yes, if you could change your PR to add that, that would be nice.

Rovanion:
Modified, hope it fits the bill!

borkdude:
Thanks!

PR #377, opened 2020-04-23
add whatsapp chat histogram example

Kah0ona:
I thought this was kinda fun, but, if it's not interesting enough as an example for this repo, dismiss this PR :-)


borkdude:
Thanks 😎

PR #400, opened 2020-04-30
Add my project - close #392

### LINKED ISSUE #392: add https://github.com/cldwalker/clj-clis to list of projects

cldwalker:
Saw the issue so figured I'd add it :) It could be useful to make a distinction in that document between projects that use babashka and libraries that are compatible with babashka

### DIFF
--- a/doc/projects.md
+++ b/doc/projects.md
@@ -197,3 +197,7 @@ FAIL in (footest) (:1)
 expected: :foox
   actual: :foo
 ```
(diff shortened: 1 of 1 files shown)


borkdude:
@cldwalker 

> It could be useful to make a distinction in that document between projects that use babashka and libraries that are compatible with babashka

Yes, that would be useful. PR for that also welcome!

Issue #388, opened 2020-04-27
write postgres test using embedded postgres


borkdude:
https://github.com/opentable/otj-pg-embedded

kolharsam:
@borkdude I'd like to help with this if that's alright.

borkdude:
@kolharsam Sure!

Right now, JDBC + Postgres support is behind a feature flag. But I think it would be nice if we could test the JDBC / Postgres functionality when running the tests in on the JVM using an embedded Postgres. Same for HSQLDB. 

So:

- add the embedded postgres driver to the test dependencies in project.clj
- same for the hsqldb driver
- implement unit tests for postgres + hsqldb (in memory) in namespace: babashka.jdbc-test
- only run it when `babashka.test-utils/jvm?` is true, so we don't have to build a native image for it (those features are not part of the standard build)

Also see:

- the test suite of https://github.com/seancorfield/next-jdbc. 
- script/postgresql-test: this test runs using a Docker container, but that doesn't work in CI

kolharsam:
@borkdude,
question regarding the unit tests that have to be written...

would they have to be as extensive as they are in `next-jdbc` or just the `basic` & `connectivity` test suites are good enough? 

I've never written tests like these and so I'd like to know what kind of tests we write for databases?

Also, will the tests be run behind the feature flags?

borkdude:
@kolharsam The test is basically already there:

https://github.com/borkdude/babashka/blob/master/test-resources/babashka/postgresql_test.clj

It's basically just a sanity check to see if this functionality still works. Take a look around in the other tests for examples.

The only difference to be able to run this with embedded postgres is the driver name: `:dbtype "embedded-postgres"` instead of `:dbtype "postgres"` (although it could be  worth trying if it just works with `postgres`). 

Here is some test code from next.jdbc that uses this embedded driver:

https://github.com/seancorfield/next-jdbc/blob/780c75954176ade9d523eb3f1bda001b5bdc0b95/test/next/jdbc/test_fixtures.clj#L98

kolharsam:
@borkdude you can track the work on https://github.com/borkdude/babashka/pull/409

Also, when I ran `lein deps` I'm not able to install all dependencies. I get an error that I could not retrieve certain resources like `edamame`. How do I rectify this?

And about `:dbtype` where is that key available?

Apologies, that I keep asking about elementary issues.

borkdude:
[...]

PR #409, opened 2020-05-02
Embedded Postgres for tests on bb

kolharsam:
See issue: #388

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,8 +12,6 @@ jobs:
     environment:
       LEIN_ROOT: "true"
       BABASHKA_PLATFORM: linux # could be used in jar name
-      BABASHKA_FEATURE_JVM: "true"
-      BABASHKA_FEATURE_POSTGRESQL: "true"
     resource_class: large
     steps:
       - checkout
@@ -40,6 +38,8 @@ jobs:
       - run:
           name: Run JVM tests
           command: |
+            echo 'export BABASHKA_FEATURE_JDBC=true' >> $BASH_ENV
+            echo 'export BABASHKA_FEATURE_POSTGRESQL=true' >> $BASH_ENV
             script/test
             script/run_lib_tests
       # - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,8 +38,8 @@ jobs:
       - run:
           name: Run JVM tests
           command: |
-            echo 'export BABASHKA_FEATURE_JDBC=true' >> $BASH_ENV
-            echo 'export BABASHKA_FEATURE_POSTGRESQL=true' >> $BASH_ENV
+            export BABASHKA_FEATURE_JDBC=true
+            export BABASHKA_FEATURE_POSTGRESQL=true
             script/test
             script/run_lib_tests
       # - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,8 @@ jobs:
     environment:
       LEIN_ROOT: "true"
       BABASHKA_PLATFORM: linux # could be used in jar name
+      BABASHKA_FEATURE_JVM: "true"
+      BABASHKA_FEATURE_POSTGRESQL: "true"
     resource_class: large
     steps:
       - checkout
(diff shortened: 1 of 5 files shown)


borkdude:
@kolharsam Can you please not use force push on branches? This makes updating my local branch more difficult. Just add more commits. They get squashed anyway.

borkdude:
Also see: https://github.com/borkdude/clj-kondo/blob/master/doc/dev.md#force-push

test/babashka/postgresql_test.clj
```diff
@@ -0,0 +1,42 @@
...
+  (println "===" (-> m :var meta :name))
+  (println))
+
+(defn jdbc-feature-flag? []
+  (= (System/getenv "BABASHKA_FEATURE_JDBC") "true"))
+
+(defn pg-feature-flag? []
+  (= (System/getenv "BABASHKA_FEATURE_POSTGRESQL") "true"))
+
+(println jdbc-feature-flag? pg-feature-flag?)
```
borkdude, line 19: You are printing two functions here. Can you remove them?

test/babashka/postgresql_test.clj
```diff
@@ -0,0 +1,42 @@
...
+                            "-e" "(require '[next.jdbc :as jdbc])"
+                            "-e" "(require '[next.jdbc.sql :as sql])"
+                            "-e" "(load-file (io/file \"test-resources\" \"babashka\" \"embedded_pg_tests.clj\"))"
+                            "-e" "(require '[embedded-pg-tests :as em-pg])"
+                            "-e" "(def db {:dbtype \"postgres\" :port 54322 :user \"postgres\" :dbname \"postgres\"})"
+                            "-e" "(em-pg/wait-for-postgres db)"
+                            "-e" "(em-pg/create-table-test db)"
+                            "-e" "(em-pg/insert-test db)"
+                            "-e" "(em-pg/query-test db)")]
+        (is (= bb-tests "true\n"))
```
borkdude, line 40: I think it would be nicer if `(tu/bb nil "-e" "(em/pg-query-test)")` etc. returned the query result and we make the assertion in the JVM on the expected result `[{:foo/foo "foo"} {:foo/foo "bar"} {:foo/foo "baz"}]`.

borkdude:
Apart from the two comments above this looks good for merge. One last thing: I think the feature flags can be enabled in the CircleCI config jvm step, so the test will run only in that step.

test/babashka/postgresql_test.clj
kolharsam: Yup, sure.

kolharsam:
Have updated according to the review suggestions. 
Hope all is good now.

.circleci/config.yml
```diff
@@ -12,6 +12,8 @@ jobs:
     environment:
       LEIN_ROOT: "true"
       BABASHKA_PLATFORM: linux # could be used in jar name
+      BABASHKA_FEATURE_JVM: "true"
```
borkdude: This is not good yet. This will make the uberjar include the postgres driver. As discussed, that's not what we want.

kolharsam: Yeah, I've corrected for those in my latest commit

.circleci/config.yml
```diff
@@ -40,6 +38,8 @@ jobs:
       - run:
           name: Run JVM tests
           command: |
+            echo 'export BABASHKA_FEATURE_JDBC=true' >> $BASH_ENV
```
borkdude, line 41: why not just:
```
export BABASHKA_FEATURE_JDBC=true
``` ?

Comment (kolharsam) [reply .circleci/config.yml:41]: just the jdbc flag is enough?

Comment (borkdude) [reply .circleci/config.yml:41]: No, I mean:

why `echo 'export ...=...' >> $BASH_ENV` instead of just `export ...=...`

Comment (kolharsam) [reply .circleci/config.yml]: Changed it.

**Comment by borkdude** [comment]
@kolharsam Seems the tests run fine in CI now:

```
lein test babashka.postgresql-test
=== postgresql-test

SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

Congrats, ready for merge. I'll wait for all the tests to finish.

kolharsam:
@borkdude thanks for all the help along the way.

PR #416, opened 2020-05-07
Adds three lines of docs to example web server

bherrmann7:
I found it confusing that the example web server only accepted a single connection.     I tried to add some documentation to call this out.


borkdude:
@bherrmann7 I would also be fine with a PR to make it multi-threaded. What do you think?

bherrmann7:
I would like that, but I'm afraid I will destroy the simplicity.      However, it is probably the more common use case.    I will what I can do to meet both constraints.

borkdude:
For now I'll just merge this.

Issue #434, opened 2020-05-13
document pod meetup presentation


borkdude:
https://twitter.com/amsclj/status/1260645359002382338

Issue #431, opened 2020-05-13
document spotify project


borkdude:
https://github.com/kolharsam/bb-spotify

Issue #430, opened 2020-05-13
document blog in Japanese about AWS


borkdude:
https://tech.toyokumo.co.jp/entry/aws_logs_babashka

Issue #428, opened 2020-05-12
Document lambdaisland hiccup script 


borkdude:
ok, sure. Feel free to add it then. This one might be interesting for folks as well

https://github.com/lambdaisland/open-source/blob/master/src/lioss/subshell.clj

like `sh/sh`, but it inherits stdin/stdout/stderr, so that the user sees in real time what the subprocess is doing, and can possibly interact with it. More like how shelling out in a bash script works.

_Originally posted by @plexus in https://github.com/borkdude/babashka/issues/229#issuecomment-627257823_

PR #440, opened 2020-05-19
Update projects and other docs

kolharsam:
This should close the following tickets:
#434, #431, #430, #428, #426, #412, #383, #381


borkdude:
Thanks! One minor issue: I'd like to sort the articles and podcast from newest to oldest. So if you could add the newest three items on top of the list, that would be nice.

PR #442, opened 2020-05-20
Update CircleCI to use Graal v19.3.2

kolharsam:


borkdude:
Thanks.

Next steps:

- [ ] update Github actions workflow as well
- [ ] Update Appveyor config

Might be nice to write a babashka script that can automatically upgrade this in the future.

PR #443, opened 2020-05-21
Add bump graal version script

kolharsam:
This was also requested in #442

### DIFF
--- a/script/bump_graal_version.clj
+++ b/script/bump_graal_version.clj
@@ -14,17 +14,25 @@
 ;; Currently we use GraalVM java8-19.3.2
 
 (ns bump-graal-version
-  (:require [clojure.string :as str]))
+  (:require [clojure.string :as str]
+            [clojure.tools.cli :as cli]))
 
 (defn display-help []
   (println (->> [""
                  "This is a script that should be run when you'd"
-                 "you'd like to bump the GraalVM version for bb."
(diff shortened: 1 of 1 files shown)


script/bump_graal_version.clj
```diff
@@ -0,0 +1,76 @@
...
+(defn is-valid-bump?
+  [version]
+  (some #(= % version) valid-bumps))
+
+(defn run-shell
+  [sed-instr file]
+  ;; the empty string is added after -i so that
+  ;; a backup file isn't generated.
+  ;; this is issue specific to macOS
+  (shell/sh "sed" "-i" "" "-e" sed-instr file))
```
borkdude, line 50: I'd prefer it if we could just use slurp + str/replace instead of shelling out. This makes it more platform independent (e.g. runnable on Windows).

kolharsam: Made the necessary changes

kolharsam:
@borkdude will the support continue to be over `java8`? If that is not the case then perhaps we should account for that as well in this script?

borkdude:
@kolharsam It would be nice if you could extend the script to support both `java8` and `java11` as well with another CLI argument.

kolharsam:
@borkdude have updated the script with java arg

borkdude:
Thanks!

PR #456, opened 2020-05-30
Add Float class for use in scripts

cldwalker:
I'd like to `bb '(Float/parseFloat "7.5")'` and was unable to. I think this is all that is needed to enable this. Cheers


borkdude:
Thanks! I believe we also need to add this class to `imports` in babashka/main.clj

PR #462, opened 2020-06-08
Disable signal handlers on AWS lambdas via env var

Chowlz:
The introduction of shutdown hooks from v0.0.81 prevents babashka from running in an AWS lambda. This PR allows disabling the hook through the same env var for disabling the pipe signal handler and renames the env variable for simplicity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -803,8 +803,8 @@ repo](https://github.com/babashka/babashka.pods).
 ## Package babashka script as a AWS Lambda
 
 AWS Lambda runtime doesn't support signals, therefore babashka has to disable
-handling of the SIGPIPE. This can be done by setting
-`BABASHKA_DISABLE_PIPE_SIGNAL_HANDLER` to `true`.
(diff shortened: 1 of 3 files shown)


borkdude:
Thanks!

Two minor remarks:

- Maybe the env var should now be called `BABASHKA_DISABLE_SIGNAL_HANDLERS`, plural?
- The documentation should also mention `SIGINT` (as it previously only disabled handling `SIGPIPE`.

Related PR: https://github.com/borkdude/babashka/pull/305

Chowlz:
> Thanks!
> 
> Two minor remarks:
> 
>     * Maybe the env var should now be called `BABASHKA_DISABLE_SIGNAL_HANDLERS`, plural?
> 
>     * The documentation should also mention `SIGINT` (as it previously only disabled handling `SIGPIPE`.
> 
> 
> Related PR: #305

Sounds good. Made the suggested changes.

PR #512, opened 2020-07-15
Compile script now follows symlinks when searching for svm.jar

### LINKED ISSUE #511: Compile script does fails to find svm.jar on SDKMAN! installed Graal
**version**
babashka v0.1.3

**platform**
macOS 10.15.5

**symptom**
`script/compile` exits prematurely with exit code of 1.

**repro**
After successfully running `script/uberjar`, I run `script/compile`.

**actual behavior**
I get the following output:
```
Downloading: Component catalog from www.graalvm.org
Processing Component: Native Image
Component Native Image (org.graalvm.native-image) is already installed.
```
and an exit code of 1.

**expected behavior**
I expect the compile to run through to completion.

**diagnosis**
I use SDKMAN! to manage my java and GraalVM installations.  It uses symlinks:
```
> ls -l $JAVA_HOME
lrwxr-xr-x  1 lee  staff  49 26 Jun 16:35 /Users/lee/.sdkman/candidates/java/current -> /Users/lee/.sdkman/candidates/java/20.1.0.r11-grl
```

The compile script does not currently follow symlinks when searching for svm.jar and therefore fails to find it in my installation.

**action**
I will follow up with a PR.

lread:
Fixes #511

### DIFF
--- a/script/compile
+++ b/script/compile
@@ -24,7 +24,7 @@ fi
 export JAVA_HOME=$GRAALVM_HOME
 export PATH=$GRAALVM_HOME/bin:$PATH
 
-SVM_JAR=$(find "$GRAALVM_HOME" | grep svm.jar)
+SVM_JAR=$(find -L "$GRAALVM_HOME" | grep svm.jar)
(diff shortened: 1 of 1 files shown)


borkdude:
Thanks!

PR #522, opened 2020-08-08
Add java class to support circle/bond

cldwalker:
Hi borkdude. I'd like to be able to use https://github.com/circleci/bond for a bb testing tool I'm writing. Is this ok? I don't know the guidelines on what classes are ok to add but I figured anything with clojure.* was fair game. 
I'm happy to add bond to libraries.md and lib_tests/. Does one just create a new directory in test-resources/lib_tests/ and copy over tests from the library?

Cheers

### DIFF
--- a/src/babashka/impl/classes.clj
+++ b/src/babashka/impl/classes.clj
@@ -80,7 +80,8 @@
                                         :parameterTypes ["java.lang.String","java.util.Locale","java.lang.ClassLoader"]}]})))
 
 (def classes
-  `{:all [clojure.lang.BigInt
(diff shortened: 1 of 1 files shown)


borkdude:
Yes, that's fine.

borkdude:
Adding to docs + tests would be most welcome. The way you described it, is correct. Do the tests need the fix for borkdude/sci#375?

cldwalker:
Thanks! Bond tests don't seem to depend on that issue as there aren't any redefs on clojure core fns

PR #524, opened 2020-08-09
Add lib tests for bond

cldwalker:
Hi borkdude.
I'm following up on #522 to add tests for bond. There is one issue. Currently a [test fails](https://github.com/cldwalker/babashka/blob/39dd76cc2e882e35f6909dfe295a2e1ed7127ddb/test-resources/lib_tests/bond/test/james.cljc#L9-L19) because bb behaves differently than clojure. When a fn is called with an incorrect number of arguments, bb throws Exception whereas clojure throws clojure.lang.ArityException. Would it be ok to change bb to match clojure's behavior? I can attempt this change in sci if you'd like

Cheers

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -55,4 +55,5 @@
                          aero {:mvn/version "1.1.6"}
                          org.clojure/data.generators {:mvn/version "1.0.0"}
                          honeysql {:mvn/version "1.0.444"}
-                         minimallist {:mvn/version "0.0.1"}}}}}
+                         minimallist {:mvn/version "0.0.1"}
+                         circleci/bond {:mvn/version "0.4.0"}}}}}
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -41,6 +41,16 @@ Test the native version:
 
     BABASHKA_TEST_ENV=native script/test
 
+## Tests for Libraries
+
+Babashka runs tests of libraries that are compatible with it through
+`script/run_lib_tests`. To add tests for a new library, do the following:
+
+* Add an entry for the library in `deps.edn` under the `:lib-tests` alias.
+* Create a directory for the library in `test-resources/lib_tests/` and copy its tests to there.
+* Add an entry in `run_all_libtests.clj` to run the added test namespaces.
+
+
 ## Build
 
 See [build.md](build.md).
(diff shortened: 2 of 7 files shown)


borkdude:
@cldwalker By design sci throws an `ex-info` that wraps the original exception. The reason is that sci also has to propagate location data that will be included in the data of the `ex-info`. You can see this in bb:

```
$ bb '(map)'
clojure.lang.ArityException: Wrong number of args (0) passed to: clojure.core/map [at line 1, column 1]
```

vs 

```
$ bb --verbose '(map)'
clojure.lang.ExceptionInfo: Wrong number of args (0) passed to: clojure.core/map [at line 1, column 1]
{:type :sci/error, :line 1, :column 1, :message "Wrong number of args (0) passed to: clojure.core/map [at line 1, column 1]"}
 at sci.impl.utils$rethrow_with_location_of_node.invokeStatic (utils.cljc:72)
    sci.impl.interpreter$eval_call.invokeStatic (interpreter.cljc:573)
    sci.impl.interpreter$interpret.invokeStatic (interpreter.cljc:613)
    sci.impl.interpreter$eval_form.invokeStatic (interpreter.cljc:645)
    sci.impl.interpreter$eval_string_STAR_.invokeStatic (interpreter.cljc:658)
    sci.core$eval_string_STAR_.invokeStatic (core.cljc:196)
    babashka.main$main$fn__17571$fn__17572.invoke (main.clj:557)
    babashka.main$main$fn__17571.invoke (main.clj:557)
    babashka.main$main.invokeStatic (main.clj:551)
    babashka.main$main.doInvoke (main.clj:440)
    clojure.lang.RestFn.applyTo (RestFn.java:137)
    clojure.core$apply.invokeStatic (core.clj:665)
    babashka.main$_main.invokeStatic (main.clj:593)
    babashka.main$_main.doInvoke (main.clj:587)
    clojure.lang.RestFn.applyTo (RestFn.java:137)
    babashka.main.main (:-1)
```

I think it might be good to just make a small tweak to the tests to account for this.

cldwalker:
> @cldwalker By design sci throws an `ex-info` that wraps the original exception. The reason is that sci also has to propagate location data that will be included in the data of the `ex-info`. You can see this in bb:

Ah. Good to know. I've tweaked the test to catch bb's exception

borkdude:
Thanks!

Issue #468, opened 2020-06-11
Support for clojure.datafy


cldwalker:
Hi. Exciting that [defprotocol support](https://github.com/borkdude/sci/issues/279) is out! I'm opening this issue to track support for clojure.datafy.

~~Soon after defprotocol support landed on master, I tried requiring `clojure.datafy` and ran into missing java classes e.g. ` Unable to resolve classname: java.lang.Iterable [at clojure/core/protocols.clj, line 34, column 6]`. Here's what I tried to remedy missing java classes:~~

EDIT:
I'm able to require clojure.datafy with [this branch](https://github.com/cldwalker/babashka/tree/datafy-2). This branch adds all the java classes that namespace required. However when I try to exercise datafy functionality via clojure.datafy/datafy or clojure.datafy/nav, I don't see the expected behavior. To reproduce what I'm seeing, run the built ./bb of the above branch against [this file](https://gist.github.com/cldwalker/4bf72c5c120f05f117508b9556969e04) e.g. `bb -f explore_data_nav.clj`. Observe that `nav-to-link` starting with [this invocation](https://gist.github.com/cldwalker/4bf72c5c120f05f117508b9556969e04#file-explore_datafy_nav-clj-L119) returns a number instead of an expected map.

It's unclear to me if I'm adding java classes in the wrong places, if there's a sci issue with defprotocol or something else. The above failing example only navigates maps. Any thoughts?

**Is your feature request related to a problem? Please describe.**
Would like to use clojure.datafy for navigating clojure data and java classes because of its first class reach in the language

**Describe the solution you'd like**
Would like require clojure.datafy to work

**Describe alternatives you've considered**
There isn't another that has the same functionality and is built into the language

**Additional context**
n/a

borkdude:
@cldwalker `:extend-via-metadata` isn't supported yet in `defprotocol`. I'll make a new sci issue for that.

borkdude:
Added `:extend-via-metadata to bb master now.

borkdude:
@cldwalker I think clojure.datafy has to be added as a built-in namespace (either in sci or bb) due to some differences of what can be datafied (e.g. a sci namespace vs a clojure.lang.Namespace). 

What would be some useful example of having this in bb?

djblue:
[...]

PR #526, opened 2020-08-11
[#468] Add clojure.datafy namespace

borkdude:


djblue:
I can't seem to require `clojure.core.protocols` so I can implement the `Datafiable` protocol.

borkdude:
@djblue I guess that will be part 2 of this exercise.

djblue:
@borkdude it works great otherwise. I was able to datafy `*ns*` and some classes 🎉

borkdude:
@djblue What kinds of objects did you want to implement `Datafiable` protocol for, just out of curiosity?

src/babashka/impl/datafy.clj
```diff
@@ -0,0 +1,46 @@
...
+(extend-protocol p/Datafiable
+  SciNamespace
+  (datafy [n]
+    (with-meta {:name (sci-ns-name n)
+                :publics (->> n (sci-ns-publics @ctx) sortmap)
+                :imports (->> n (sci-ns-imports @ctx) sortmap)
+                :interns (->> n (sci-ns-interns @ctx) sortmap)}
+      (meta n))))
+
+(def datafy-ns (create-ns 'clojure.data nil))
```
cldwalker, line 42: Should this be 'clojure.datafy ?

borkdude: Thanks, fixed now.

Issue #565, opened 2020-09-04
version-clj fails when included via "use"


lread:
**version**
v0.2.0

**platform**
macOS v10.15.6 
openjdk v11.0.8

**problem**
When including `version-clj.core` via `use` we get a `StackOverflowError`.

**repro**
From an empty directory so you don't accidentally pick up extra deps from some project, issue the following shell commands:
```
export BABASHKA_CLASSPATH=$(clojure -Spath -Sdeps '{:deps {version-clj {:mvn/version "0.1.2"}}}')
bb -e "(use 'version-clj.core) (version->seq \"1.0.0-SNAPSHOT\")" 
```

**actual behavior**
Ouput is:
```
----- Error --------------------------------------------------------------------
Type:     java.lang.StackOverflowError
Location: version_clj/split.clj:61:3

----- Context ------------------------------------------------------------------
57: ;; ## Split Points
58: 
59: (defn- split
60:   [v s]
61:   (if (fn? v)
      ^--- 
62:     (v s)
63:     (string/split s v)))
64: 
65: ;; ## Predefined Split Points
66: 

----- Locals -------------------------------------------------------------------
v: "1.0.0-SNAPSHOT"
s: #"\."

----- Stack trace --------------------------------------------------------------
version-clj.split/split                - version_clj/split.clj:59:8
version-clj.split/split                - version_clj/split.clj:63:5
version-clj.split/split                - version_clj/split.clj:59:8
version-clj.split/first-split-at-point - version_clj/split.clj:106:15
version-clj.split/first-split-at-point - version_clj/split.clj:102:8
version-clj.split                      - version_clj/split.clj:132:20
version-clj.core/version->seq          - version_clj/split.clj:7:6
version-clj.split                      - version_clj/split.clj:127:16
version-clj.core/version->seq          - version_clj/split.clj:7:6
user                                   - <expr>:1:25
```

**expected behavior**
If we employ `require` instead of `use`:
```
bb -e "(require '[version-clj.core :as ver]) (ver/version->seq \"1.0.0-SNAPSHOT\")"
```
We get an expected output:
```
[(1 0 0) ["snapshot"]]
```

**diagnosis**
In a Slack chat, borkdude guessed that:
> split is private, but bb currently doesn't honor private

**note to fixer**
Also from chat with borkdude on Slack:
> What I usually do for libraries like this is copy the tests to test-resources and add it to the lib tests script
> so they will be tested on CI
> like a smoke test

[...]

PR #568, opened 2020-09-09
Document version-clj as a library that works with babashka

lread:
The `use` syntax relies on resolution of #565


borkdude:
@lread Thanks!

borkdude:
@lread Maybe avoid the usage of `use` in the example since this doesn't work with version 0.2.0 or below?

lread:
I prefer not using `use` myself but adopted that because that's what version-clj docs README examples used and thought folks might follow that lead.

I think you are right, I'll change from `use` to `require`.

PR #598, opened 2020-09-25
Adding the custom version support in command line args

pradeepbishnoi:
Install the custom version or latest version of babashka using command line args

### DIFF
--- a/install
+++ b/install
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 version=""
-default_install_dir="/tmp/bb"
+default_install_dir="/usr/local/bin"
 install_dir=$default_install_dir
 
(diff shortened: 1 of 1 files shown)


borkdude:
Thanks. Two questions:
- Why the change for /tmp/bb instead of /usr/local/bin?
- Does it still support the old way, i.e. only passing a directory name without flags?

pradeepbishnoi:
Resolved the default path, that tmp path was used while local testing.

borkdude:
Thanks. The only thing left is to fix compatibility with the old way of using the script: `install <dir>` .

borkdude:
Thanks a lot.

PR #626, opened 2020-10-30
Let image_viewer example take options for port and dir

rwstauner:
and don't error when the browser requests /favicon.ico


rwstauner:
This thing is great!  Thanks!

PR #638, opened 2020-11-11
Add feature/oracledb

holyjak:
NOTE: Using the Oracle driver directly requires some GraalVM settings
while the helidon wrapper provides those and thus works out of the box.


borkdude:
Thanks!

PR #642, opened 2020-11-12
next.jdbc: Add additional functions and ns

holyjak:


holyjak:
BEWARE: It seems these changes considerably increase compilation time and memory usage, leading up to out of memory errors (on my 16GB OSX, with java consuming over 5GB). So it is not a good idea to merge until the cause is found and removed.

borkdude:
@holyjak Maybe you could compare to https://github.com/babashka/babashka-sql-pods/blob/05c5d9512713d834c544e2d6c592d4988ed454de/src/pod/babashka/sql.clj#L100-L106.

In those pods it seems to work well. Could also be a version difference in next.jdbc.

holyjak:
Removed the prepare namespace, as it leads to an explosion in compile time and memory usage.

holyjak:
> In those pods it seems to work well. Could also be a version difference in next.jdbc.

What "it" do you mean? I had problems with `next.jdbc.prepare` but I do not see that ns in the pod either?

borkdude:
@holyjak I'm not sure what this is about anymore, since it's a while ago.

holyjak:
Plan of action:

1. add these vars first to the https://github.com/babashka/babashka-sql-pods
2. write real tests to verify if they work correctly and to see what impact they have on the size
3. if ok, then they can be added here

arichiardi:
Nice this is a nice one to have for us as well.

borkdude:
I'll close this PR and will re-open once some will work on this again.

PR #649, opened 2020-11-17
Add the missing class from java.io

agilecreativity:
I have use-case that required `java.io.PipedInputStream` and `java.io.PipedOutputStream` in code which is missing from Babashka.

This PR attempt to add these two classes to Babashka.


borkdude:
I will first merge this to a temporary branch to inspect the increase in binary size.

borkdude:
Merged to master now.

PR #672, opened 2020-12-08
Add features Ring, Reitit, Selmer 

kloimhardt:
Ring and Reitit needed patches to be Graal compilable without blowing up the binary size.
Ring was also patched to accept the additional key `:io-resource-fn` in

```
(http/run-server
    (wrap-defaults
      handler
      (assoc api-defaults :static {:resources "public"
                                   :io-resource-fn io/resource}))
    {:port port})
```


deps.edn
```diff
@@ -20,6 +21,13 @@
         clj-commons/clj-yaml {:mvn/version "0.7.2"}
         com.cognitect/transit-clj {:mvn/version "1.0.324"}
         nrepl/bencode {:mvn/version "1.1.0"}
+        ring/ring-core {:mvn/version "1.8.1"}
+        ring/ring-defaults {:mvn/version "0.3.2"}
+        ring-webjars {:mvn/version"0.2.0"}
+        metosin/reitit {:mvn/version "0.5.5"}
+        metosin/muuntaja {:mvn/version "0.6.7"}
+        metosin/ring-http-response {:mvn/version "0.9.1"}
+        selmer {:mvn/version "1.12.29"}}
```
borkdude, line 30: I think there's a typo here which makes the tests fail.

kloimhardt: Thank you for the hint to this all too obvious mistake of mine.

kloimhardt: fixed it

PR #744, opened 2021-03-02
Don't leave the user stranded in install_dir

guss77:
After the installation completes, the user would be CDed into `$install_dir` regardless of where they started from - this is bad practice.

Its better to run actual operations (i.e. other than parsing input and setting local variables) in a subshell, so when it exists - the user is back to where they started. shell options and exit codes will be translated automatically across the subshell

### DIFF
--- a/install
+++ b/install
@@ -66,6 +66,7 @@ esac
 
 download_url="https://github.com/babashka/babashka/releases/download/v$version/babashka-$version-$platform-amd64.zip"
 
+(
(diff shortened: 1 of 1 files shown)


borkdude:
@guss77 When is this a problem? Maybe only when you `source` the script? In other cases the script is already its own subshell?

guss77:
Hmm.. can't repro the problem I had now. I'll re-review and re-open if needed.

PR #755, opened 2021-03-13
Augment proxy to fulfill custom maps interfaces

wilkerlucio:


src/babashka/impl/proxy.clj
```diff
@@ -8,22 +8,45 @@
...
-      (seq [] ((method-or-bust methods 'seq) this)))
+      (seq [] ((method-or-bust methods 'seq) this))
+
+      (equiv [other]
+        (if-let [m (get methods 'equiv)]
+          (m this other)
+          (proxy-super equiv other)))
+      (empty [] ((method-or-bust methods 'empty) this))
+
+      (meta [] ((method-or methods 'meta nil) this))
```
wilkerlucio, line 40: although those are not part of the core APersistentMap, custom maps are usually expected to support meta, the way I did here is so the user can extend it, otherwise just make it transparent (meta returns nil, with-meta returns self)

PR #760, opened 2021-03-24
Update README.md to use sudo in install

time2die:
you must use sudo to install babashka
image


borkdude:
This is already explained here:

https://github.com/babashka/babashka#installer-script

Suggesting sudo by default might not be want people want, so it's best if they followed the install instructions.

PR #777, opened 2021-04-06
Use get-classpath when build an uberjar

arichiardi:
It was previously using the classpath coming from the parameters/env. We need to use
cp/get-classpath for taking into consideration the classpath coming from bb.edn.


borkdude:
Thanks :)

PR #803, opened 2021-04-25
Adds Expound to projects.md

bhb:
This duplicates the example in [spartan.spec](https://github.com/borkdude/spartan.spec#example), but I figure that in some cases, people will check this list of projects before they think to check the READMEs of individual projects.


borkdude:
Thanks!

PR #828, opened 2021-05-09
Build BABASHKA_STATIC with musl C library

thiagokokada:
Based on https://github.com/babashka/pod-babashka-aws/pull/37.

This is the recommended way to build static binaries with GraalVM by the documentation and multiple issues on GitHub. See:

- https://www.graalvm.org/reference-manual/native-image/StaticImages/
- oracle/graal#571 (comment)

The reason that building a statically binary with glibc is complicated is explained in the glib documentation:

https://sourceware.org/glibc/wiki/FAQ#Even_statically_linked_programs_need_some_shared_libraries_which_is_not_acceptable_for_me.__What_can_I_do.3F

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,13 @@
 FROM clojure:lein-2.9.1 AS BASE
 
+ENV DEBIAN_FRONTEND=noninteractive
 RUN apt update
-RUN apt install --no-install-recommends -yy curl unzip build-essential zlib1g-dev
+RUN apt install --no-install-recommends -yy curl unzip build-essential zlib1g-dev sudo
 WORKDIR "/opt"
 RUN curl -sLO https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-21.0.0/graalvm-ce-java11-linux-amd64-21.0.0.tar.gz
 RUN tar -xzf graalvm-ce-java11-linux-amd64-21.0.0.tar.gz
 
-ARG BABASHKA_XMX="-J-Xmx3g"
+ARG BABASHKA_XMX="-J-Xmx4500m"
 
 ENV GRAALVM_HOME="/opt/graalvm-ce-java11-21.0.0"
 ENV JAVA_HOME="/opt/graalvm-ce-java11-21.0.0/bin"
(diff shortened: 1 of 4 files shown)


thiagokokada:
Tested with `docker build -t babashka . --build-arg BABASHKA_STATIC=true`.

Compared to PR https://github.com/babashka/pod-babashka-aws/pull/37, I did some changes:
- Instead of hardcoding `/usr/lib/gcc/x86_64-linux-gnu/8`, I decided to find directory correctly, otherwise if the GCC version changes this build will break (building with `Dockerfile` uses GCC 6, so hardcoding `/usr/lib/gcc/x86_64-linux-gnu/8` was breaking the build).
- Needed to add `apt-get -y`, otherwise the installation would get stuck in `Dockerfile`.

CC @borkdude @lispyclouds .

Dockerfile
```diff
@@ -1,12 +1,13 @@
...
+ENV DEBIAN_FRONTEND=noninteractive
 RUN apt update
-RUN apt install --no-install-recommends -yy curl unzip build-essential zlib1g-dev
+RUN apt install --no-install-recommends -yy curl unzip build-essential zlib1g-dev sudo
 WORKDIR "/opt"
 RUN curl -sLO https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-21.0.0/graalvm-ce-java11-linux-amd64-21.0.0.tar.gz
 RUN tar -xzf graalvm-ce-java11-linux-amd64-21.0.0.tar.gz
 
-ARG BABASHKA_XMX="-J-Xmx3g"
+ARG BABASHKA_XMX="-J-Xmx4500m"
```
borkdude, line 10: Why this change?

thiagokokada: With only 3GiB the build doesn't work anymore, it needs more memory.

thiagokokada: I set to `-J-Xmx4500m` to mirror what is set here: https://github.com/thiagokokada/babashka/blob/6fba4f968300318e00cc29278f49bfc17266ca22/script/compile

borkdude: ok

PR #842, opened 2021-05-14
Improve install script

thiagokokada:
- Add `--checksum` option. This will allow to check the SHA256 sum of the download, making sure that we are downloading the correct binary. By default, there is no check, to keep this script backwards compatible (e.g.: if someone doesn't use `--checksum` option, they don't need to have `sha256sum` installed)
- Add `--static` option. This will, on Linux systems, download the static release instead of the dynamic one. On macOS it exit with error
- Some small refactorings

### DIFF
--- a/install
+++ b/install
@@ -3,16 +3,17 @@
 set -euo pipefail
 
 version=""
+checksum=""
 default_install_dir="/usr/local/bin"
-install_dir=$default_install_dir
+install_dir="$default_install_dir"
 default_download_dir="/tmp"
-download_dir=$default_download_dir
+download_dir="$default_download_dir"
 
 print_help() {
     echo "Installs latest (or specific) version of babashka. Installation directory defaults to /usr/local/bin."
     echo -e
     echo "Usage:"
-    echo "install [--dir <dir>] [--download-dir <download-dir>] [--version <version>]"
+    echo "install [--dir <dir>] [--download-dir <download-dir>] [--version <version>] [--checksum <checksum>]"
     echo -e
     echo "Defaults:"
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -25,14 +33,10 @@ if [[ $# -eq 1 ]]; then
    install_dir=${1:-}
 fi
 
-while [[ $# -gt 1 ]]
+while [[ $# -gt 0 ]]
 do
     key="$1"
-    if [[ -z "${2:-}" ]]; then
-        print_help
-    fi
```
thiagokokada, line 33: Those changes were necessary to support boolean options like `--static`. As a bonus, `--help` now works too (before it just ignored and started download with the default options).

install
```diff
@@ -3,20 +3,28 @@
...
     echo -e
     echo "Usage:"
-    echo "install [--dir <dir>] [--download-dir <download-dir>] [--version <version>]"
+    echo "install [--dir <dir>] [--download-dir <download-dir>] [--version <version>] [--checksum <checksum>] [--static]"
     echo -e
     echo "Defaults:"
     echo " * Installation directory: ${default_install_dir}"
     echo " * Download directory: ${default_download_dir}"
+    if [[ -z "$checksum" ]]; then
+        echo " * Checksum: no"
```
borkdude, line 23: I don't believe we need this logging, do we?

thiagokokada: @borkdude Done.

borkdude: Oh sorry, I misread, this is the help output. 🤦

borkdude: Feel free to add it back, my bad.

Issue #845, opened 2021-05-15
clojure.java.browse/browse-url not working in NixOS


thiagokokada:
**version**

babashka v0.4.1

**platform**

NixOS

**problem**

In Babashka `clojure.java.browse/browse-url` doesn't work:

```
Babashka v0.4.1 REPL.
Use :repl/quit or :repl/exit to quit the REPL.
Clojure rocks, Bash reaches.

user=> (require '[clojure.java.browse :as b])
nil
user=> (b/browse-url "https://www.google.com")
java.io.IOException: Cannot run program "/usr/bin/xdg-open": error=2, No such file or directory [at <repl>:2:1]2:1]
```

It works in Clojure though:

```
Clojure 1.10.1
user=> (require '[clojure.java.browse :as b])
nil
user=> (b/browse-url "https://www.google.com")
true
```

In NixOS, `xdg-open` is not available in `/usr/bin/xdg-open`, it is in `/run/current-system/sw/bin/xdg-open`. Clojure does the correct thing and tries to auto-detect the binary location: https://github.com/clojure/clojure/blob/ef00c7cffad10e6104d333f7f71a29163b06cd0a/src/clj/clojure/java/browse.clj#L21-L27.

I don't know why Babashka behavior is different here, but the correct behavior would also be to try to auto-detect the location using `which` (or even `babashka.fs/which`).

**repro**

On any minimalistic Linux distro (like Alpine), you can try to put the `xdg-open` binary in some other place in the path and call `(clojure.java.browse/browse-url "https://www.google.com")`.

**expected behavior**

The same one as Clojure.

thiagokokada:
Seeing the implementation of `babashka.impl.clojure.java.browse`, it is clear what is the problem: https://github.com/babashka/babashka/blob/master/src/babashka/impl/clojure/java/browse.clj#L20

I can provide a PR using `babashka.fs/which` to fix this issue. WDYT @borkdude ?

thiagokokada:
Supporting `*open-url-script*` (https://github.com/clojure/clojure/blob/ef00c7cffad10e6104d333f7f71a29163b06cd0a/src/clj/clojure/java/browse.clj#L40) also seems a good idea, since some people in Linux may have a custom script that is used in place of `xdg-open`.

thiagokokada:
Closed by #846.

PR #846, opened 2021-05-15
Improve `clojure.java.browse`

thiagokokada:
# Context

This PR add two improvements to `babashka.impl.clojure.java.browse`, fixing issue https://github.com/babashka/babashka/issues/845.

## Remove `xdg-open` hardcoded path in `babashka.impl.clojure.java.browse`

Binaries in Linux are not guarantee to be located on `/usr/bin`.

For example, NixOS the binaries are located in `/run/current-system/sw/bin/xdg-open` (that is actually just a symlink to `/nix/store`). The distro may also not have `xdg-open` installed at `/`, instead the user may put it somewhere else like `$HOME/bin`.

This commit fixes this by using not hardcoding the PATH of `xdg-open`, instead relying on shell path to search it.

## Implement `clojure.java.browse/*open-url-script*`

This allows usage of a custom URL script, like:

```clojure
(binding [*open-url-script* (atom "my-browse-url-script")]
  (browse-url "https://google.com"))
```

`*open-url-script*` is an atom just to keep compatibility with Clojure.

Described here: <https://clojuredocs.org/clojure.java.browse/*open-url-script*>

### DIFF
--- a/src/babashka/impl/clojure/java/browse.clj
+++ b/src/babashka/impl/clojure/java/browse.clj
@@ -1,7 +1,10 @@
 (ns babashka.impl.clojure.java.browse
   {:no-doc true}
   (:require [clojure.java.shell :refer [sh]]
-            [clojure.string :as str]))
+            [clojure.string :as str]
+            [sci.core :as sci]))
+
+(def open-url-script (sci/new-dynamic-var '*open-url-script* (atom nil)))
 
 (def os
   (let [os-name (System/getProperty "os.name")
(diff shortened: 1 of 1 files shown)


thiagokokada:
Tested with this binary:

[babashka-0.4.2-SNAPSHOT-linux-amd64-static.tar.gz](https://github.com/babashka/babashka/files/6481841/babashka-0.4.2-SNAPSHOT-linux-amd64-static.tar.gz)

src/babashka/impl/clojure/java/browse.clj
```diff
@@ -1,7 +1,10 @@
 (ns babashka.impl.clojure.java.browse
   {:no-doc true}
   (:require [clojure.java.shell :refer [sh]]
-            [clojure.string :as str]))
+            [clojure.string :as str]
+            [sci.core :as sci]))
+
+(def open-url-script (sci/new-dynamic-var '*open-url-script* (atom nil)))
```
thiagokokada, line 7: To actually have the exactly same behavior as Clojure, this atom needs to start with value `:uninitialized` and be initialized at the first call of `browse-url`.

thiagokokada: In commit f9f7582bd1b0202cca91cc2e8ff4dbf9fe970beb I set the initial value to `*open-url-script*` atom to `:uninitialized` to make it the same as Clojure. I don't see a reason to set this value at the first `clojure.java.browse/browse-url` call, though.

thiagokokada: Now that I think about it, I don't see a single reason on why someone would want to check if this value is set to `:uninitialized` first, since the only reason this dynamic var seems to exist is for initialization performance (not an issue here) and for rebinding (this last one it doesn't matter what is the initial value of this variable is).

I will remove commit f9f7582bd1b0202cca91cc2e8ff4dbf9fe970beb and revert this to the initial value of `nil`.

Issue #730, opened 2021-02-11
Improvement to throw ExceptionInfo output 


GreshamDanielStephens:
Currently when an ExceptionInfo is thrown, the data map details are lost in the output from the error. It would be nice to have the data presented.
To recreate:
```
(def some-map {:a 1 :b 2 :c 3})
(throw (ex-info "ABC" some-map))
```

Current output:
```
----- Error --------------------------------------------------------------------
Type:     clojure.lang.ExceptionInfo
Message:  ABC
Location: path/x/y/z.clj----- Context ------------------------------------------------------------------
245: 
246: 
247: 
248: 
249: (throw (ex-info "ABC" some-map))
     ^--- ABC
```

Improvement would be to add data into the output as:
```
----- Error --------------------------------------------------------------------
Type:     clojure.lang.ExceptionInfo
Message:  ABC
Data: XYZ
```

GreshamDanielStephens:
I started having a look at this but I think it's slightly more involved than I expected.
I set up a new test, doing:
```
(try (tu/bb nil "
(defn foo [] (throw (ex-info \"ex-info message\" {:some :data})))
(foo)")
                    (catch Exception e (ex-message e)))
```

Interestingly under the hood this produces an ExceptionInfo with the data:
```
{:type :sci/error, :line 2, :column 14, :message \"ex-info message\", :sci.impl/callstack #object[clojure.lang.Delay 0x7fe33ae0 {:status :ready, :val ((#'user/foo))}], :file \"<expr>\", :locals {}, :some :data}
```

So as can be seen the ex-data from the code actually gets merged in to the sci exception ex-data at the top level.
I thought that looked a bit worrying so I tried doing `(throw (ex-info \"ex-info message\" {:column :data}))` instead and sure enough we get a weird exception, which is actually a babashka bug because column gets overwritten in the sci exception with a non-int value.
```
class clojure.lang.Keyword cannot be cast to class java.lang.Number (clojure.lang.Keyword is in unnamed module of loader 'app'; java.lang.Number is in module java.base of loader 'bootstrap')
```

I haven't got too much further than that but my initial assumption from the `:type :sci/error` is that this will first require a bug fix in sci, to nest the 'inherited' ex-data down one level I think, that sounds like it might be a breaking change but I will start looking at where that fix might live when I next get a chance.

borkdude:
[...]

PR #854, opened 2021-05-18
feat: print data of uncaught ExceptionInfo to output

rng-dynamics:
I hope I won't step on anyone's feet, but I have written the two lines of code for #730. Is there anything else needed for this issue?

### DIFF
--- a/src/babashka/impl/error_handler.clj
+++ b/src/babashka/impl/error_handler.clj
@@ -100,6 +100,8 @@
                                 (.. e getClass getName)))
           (when-let [m (.getMessage e)]
             (println (str "Message:  " m)))
+          (when-let [cdt (ex-data (.getCause e))]
+            (println (str "Data:     " cdt)))
           (let [{:keys [:file :line :column]} d]
             (when line
               (println (str "Location: "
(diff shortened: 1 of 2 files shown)


borkdude:
@rng-dynamics Thanks! It might be better to use `prn` or pprint to print the map itself, to preserve the ability to parse it. 
A test would also be nice.

rng-dynamics:
Two Points:

- The output of bababshka is now as follows:
```
$ ./bb -e "(let [d {:zero 0 :one 1}] (throw (ex-info \"some msg\" d)))"
Message:  some msg
Data:     {:zero 0, :one 1}
Location: <expr>:1:27

----- Context ------------------------------------------------------------------
1: (let [d {:zero 0 :one 1}] (throw (ex-info "some msg" d)))
                             ^--- some msg

----- Locals -------------------------------------------------------------------
d: {:zero 0, :one 1}

```
The data is printed two times, now. Once in the Error section and once in the Locals section. Is that how you imagined it?

- I didn't write much real-world Clojure before. If there is anything unconventional in my code, then please let me know.

rng-dynamics:
@borkdude thanks for merging! Awesome.

PR #861, opened 2021-05-21
Use shasum instead of sha256sum

thiagokokada:
`sha256sum` is not available in macOS by default.

See: https://github.com/ESGF/esg-search/issues/84#issuecomment-213694850

However, `shasum` seems to be avaialble, so use that instead.

P.S.: not actually tested at macOS since I don't have one.

### DIFF
--- a/install
+++ b/install
@@ -118,9 +118,9 @@ rm -rf "$filename"
 rm -rf "bb"
 curl -o "$filename" -sL "$download_url"
 if [[ -n "$checksum" ]]; then
-    if ! echo "$checksum $filename" | sha256sum --check --status; then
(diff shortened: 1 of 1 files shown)


thiagokokada:
Not sure if I should use `[skip ci]` here, but I set it since I think this script has no testings.

borkdude:
Yes, [skip ci] is ok. 

I do have `sha256sum` on my system, can't remember if I installed it myself. Is there a big difference?

thiagokokada:
> I do have sha256sum on my system, can't remember if I installed it myself. Is there a big difference?

Except for the fact that it doesn't seem installed by default, no, it seems to work mostly the same (except that the filename needs to start with `*`).

PR #867, opened 2021-05-25
Improve script/setup-musl

thiagokokada:
I was trying to understand why we needed that hack to copy `libz.a` to the GCC library path even if GraalVM is supposed to set the linker correctly if we had `musl-gcc` at our PATH.

I finally figured the issue. As you can see in the code below, `zlib` sets `libdir` to the directory passed on `--prefix` and appends `/lib`:

https://github.com/madler/zlib/blob/v1.2.11/configure#L76

This would work fine if `/usr/lib` was the expected location for libraries (however the script was still wrong, since `--prefix` would need to be set to `/usr` for it to work correctly), however on Debian each library has the following format depending on the architecture/compiler:

- `/usr/lib/x86_64-linux-musl`
- `/usr/lib/x86_64-linux-gcc`

So passing `/usr/lib/x86_64-linux-musl` to the `--prefix` parameter on `zlib` results on `libz.a` installed on:

- `/usr/lib/x86_64-linux-musl/lib`

That is clearly wrong. Sadly there is no way to fix this without patching `configure` script on `zlib`, so let's change `--prefix` to
`/usr/local`, that is the correct place for non-distro programs/libraries anyway, and symbolic link `/usr/local/lib/libz.a` to
the correct path (`/usr/lib/x86_64-linux-musl`). This works and it is quite a clean solution.

### DIFF
--- a/script/setup-musl
+++ b/script/setup-musl
@@ -47,16 +47,10 @@ arch=${BABASHKA_ARCH:-"x86_64"}
 echo "ARCH: $arch"
 
 cd "zlib-${ZLIB_VERSION}"
-CC=musl-gcc ./configure --static --prefix=/usr/lib/$arch-linux-musl/
+CC=musl-gcc ./configure --static --prefix="/usr/local"
 make CC=musl-gcc
-sudo make install
-export CC=gcc
+make install
 cd ..
 
-# depending on GCC version, we will have different directories here.
-# for example, for GCC 6.3.0 we will have:
-# - /usr/lib/gcc/x86_64-linux-gnu/6
-# - /usr/lib/gcc/x86_64-linux-gnu/6.3.0
-for dest_dir in /usr/lib/gcc/$arch-linux-gnu/*; do
-    sudo cp -f /usr/lib/$arch-linux-musl/lib/libz.a "$dest_dir"
-done
+# Add a symbolic link pointing libz to the correct place so ldd can find it
+ln -sf "/usr/local/lib/libz.a" "/usr/lib/$arch-linux-musl/libz.a"
--- a/script/setup-musl
+++ b/script/setup-musl
@@ -52,5 +52,5 @@ make CC=musl-gcc
 make install
 cd ..
 
-# Add a symbolic link pointing libz to the correct place so ldd can find it
-ln -sf "/usr/local/lib/libz.a" "/usr/lib/$arch-linux-musl/libz.a"
+# Install libz.a in the correct place so ldd can find it
+install -Dm644 "/usr/local/lib/libz.a" "/usr/lib/$arch-linux-musl/libz.a"


script/setup-musl
```diff
@@ -47,16 +47,10 @@ arch=${BABASHKA_ARCH:-"x86_64"}
 echo "ARCH: $arch"
 
 cd "zlib-${ZLIB_VERSION}"
-CC=musl-gcc ./configure --static --prefix=/usr/lib/$arch-linux-musl/
+CC=musl-gcc ./configure --static --prefix="/usr/local"
 make CC=musl-gcc
-sudo make install
```
thiagokokada, line 52: Removed because this whole script run as root.

script/setup-musl
```diff
@@ -47,16 +47,10 @@ arch=${BABASHKA_ARCH:-"x86_64"}
...
-export CC=gcc
```
thiagokokada, line 53: Removed because we don't build anything else on this script, and also because we use explicitly `CC` while building `zlib`, so we don't ever set this to another value.

borkdude:
@lispyclouds Can you review this?

script/setup-musl
```diff
@@ -47,16 +47,10 @@ arch=${BABASHKA_ARCH:-"x86_64"}
...
 
-# depending on GCC version, we will have different directories here.
-# for example, for GCC 6.3.0 we will have:
-# - /usr/lib/gcc/x86_64-linux-gnu/6
-# - /usr/lib/gcc/x86_64-linux-gnu/6.3.0
-for dest_dir in /usr/lib/gcc/$arch-linux-gnu/*; do
-    sudo cp -f /usr/lib/$arch-linux-musl/lib/libz.a "$dest_dir"
-done
+# Add a symbolic link pointing libz to the correct place so ldd can find it
+ln -sf "/usr/local/lib/libz.a" "/usr/lib/$arch-linux-musl/libz.a"
```
lispyclouds, line 56: @thiagokokada this looks good to me, i was thinking we can just copy this for simplicity? anyways its running in an ephemeral CI and I guess not much people would run it on their machines?

lispyclouds:
@borkdude if you want I can follow up on this and merge this too.

borkdude:
ok!

script/setup-musl
thiagokokada: Yeah, I can copy this using `cp` or `install`, but any technical reason to prefer copying instead of linking?

And yeah, this is not used on local machines generally (or even recommended BTW), however I still think that at least putting this file on the correct place can avoid issues on the future (maybe we eventually need to use some library that depends on `libz.a` from Glibc, however thanks to the previous hack this would break and need some extra workaround).

lispyclouds: not much tech reasons for me though, im a bit wary of symlinks in general for these things, specially with permissions and who's removing what from either end. mostly end up copying to keep things isolated at the cost of bit more space, specially when its a throw away CI machine

thiagokokada: @lispyclouds Done. Decided to went with the `install` route since this will ensure that the correct permissions are set and the destination directories exist.

lispyclouds (approved):

PR #870, opened 2021-05-28
Improvements to install script

thiagokokada:
- Use `mktemp -d` to generate a temporary directory, only if user doesn't use `--download-dir`
- Clean-up the temporary directory using `trap`. Remove other `rm` calls from code since they're not necessary
  anymore
- Run download in a subshell, so after finishing it the script is still on the directory where the user ran
  the script. Fix `./install --dir .` usage
- Fix backup of old Babashka install

### DIFF
--- a/install
+++ b/install
@@ -7,8 +7,7 @@ checksum=""
 static_binary="false"
 default_install_dir="/usr/local/bin"
 install_dir="$default_install_dir"
-default_download_dir=${TMPDIR:-"/tmp"}
-download_dir="$default_download_dir"
+download_dir=""
 
 print_help() {
     echo "Installs latest (or specific) version of babashka. Installation directory defaults to /usr/local/bin."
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -7,9 +7,15 @@ checksum=""
 static_binary="false"
 default_install_dir="/usr/local/bin"
 install_dir="$default_install_dir"
-default_download_dir=${TMPDIR:-"/tmp"}
+default_download_dir="$(mktemp -d)"
```
borkdude, line 10: What if `mktemp` is not there, should we account for that?

borkdude: Maybe prevent maybe a temp dir if the user provides `--download-dir` ?

thiagokokada: `mktemp` should always be available (even if it is not defined in POSIX, macOS, GNU coreutils, busybox and toybox includes implementation of `mktemp`).

But yeah, better not to call it when `--download-dir` is passed.

thiagokokada: @borkdude Done, WDYT?

borkdude:
Thanks! Next time, consider using [skip ci] so we don't stress CircleCI too much :)

littleli (approved):

Issue #235, opened 2020-01-14
Port tests to Windows


borkdude:
Windows builds are done using Appveyor:

https://ci.appveyor.com/project/borkdude/babashka

Some tests are marked with `^:windows` metadata which are tested on Appveyor.

Many tests are failing for trivial reasons:

```
actual: (not (= "process 2\n" "process 2\r\n")
```

We already have a predicate `babashka.main/windows?` that determines if babashka is running in Windows, which can be used.

There is now also a `test-utils/normalize` function which is used to replace `\r\n` to `\n` when tests are running in Windows.

Some tests might not be suited for running on Windows, which can be handled by wrapping them in a conditional: `(when-not main/windows? ...)`.

Multiple smaller PRs to the windows-tests branch, one for each test are welcome.

borkdude:
@kolharsam If you want, you can help fix the Windows tests.

I'll update the issue with some instructions.

bobisageek:
Is this still 'available'? I started messing around with some of the tests, so I could put in a PR that addresses about 10 tests (on the JVM env), but the windows-tests branch is also well behind master, so do you think it would make more sense to merge in from master first?

borkdude:
Hi @bobisageek, thanks for your interest in adding/improving Windows tests. I now brought the `windows-tests` branch up to date, but lots of tests are failing. A PR and fixes to this branch are more than welcome. I was thinking: perhaps we can add metadata to tests so only a selection of tests run on Windows, so we can merge the branch into master and work more gradually on it.

borkdude:
I now added a test selector called `:windows` and flagged one test with `:windows` metadata. 

This is running here: https://ci.appveyor.com/project/borkdude/babashka/builds/39621659

If all tests still pass on the other platforms, I will just merge this branch so we can continue on master.

borkdude:
Merged. Let's continue on the master branch by adding `^:windows` gradually to more tests.

bobisageek:
[...]

PR #888, opened 2021-06-16
add windows selector to a couple of tests (#235)

bobisageek:
In the spirit of starting small, this change just adds the windows selector to a few tests that pass with no other changes needed.


borkdude:
Thanks, looks good

PR #894, opened 2021-06-18
add more windows testing (#235)

bobisageek:
- just handling line endings and escaping backslashes in file paths

this almost closes the work on porting the `b.main-test` namespace - I think there are only 5-10 tests left, and some of them rely on Unix-y things


borkdude:
@bobisageek great! perhaps when all have been reviewed, we remove `:windows` and introduce `:skip-windows` for the remaining tests that are too unixy.

bobisageek:
Yeah, that's what I was thinking as well... I've passed the halfway point in my local testing, so I'm going to start working on that, because now it'll be quicker to mark the ones to skip. :)

PR #898, opened 2021-06-19
make tests run on windows as the default (#235)

bobisageek:
this is kind of a big PR in my mind, so I apologize for that, but a large amount of the change is 'swapping' :windows and :skip-windows

- change from selecting tests to run to selecting tests to skip (remove all :windows tags, add
  :skip-windows tag to tests that don't currently work on windows)
- handfuls of calls to `normalize` and `escape-file-paths` to handle platform differences
- split `b.bb-edn-test/task-test` to make most of the tests run on windows, and exclude just a couple of Unix-y tests

I think the difference between Windows tests and Linux tests is now something like 10-15 assertions, but one of the big ones is the pods test, so if it's ok with you, I'd like to take some time and maybe play around with making the pod test work on Windows JVM (it works as a native test). I'd also like to (maybe) write a couple tests aimed specifically at Windows (to account for the existing Unix-y tests), and add a default test selector to skip ":windows-only" or something like that... what do you think?

### DIFF
--- a/script/test.bat
+++ b/script/test.bat
@@ -11,16 +11,32 @@ set PATH=%GRAALVM_HOME%\bin;%PATH%
 set BABASHKA_PRELOADS=
 set BABASHKA_CLASSPATH=
 set BABASHKA_PRELOADS_TEST=
+set BABASHKA_CLASSPATH_TEST=
+set BABASHKA_POD_TEST=
+set BABASHKA_SOCKET_REPL_TEST=
 
 echo "running tests part 1"
-call lein do clean, test :windows
+call lein do clean, test :windows || exit /B 1
 
 set BABASHKA_PRELOADS=(defn __bb__foo [] "foo") (defn __bb__bar [] "bar")
 set BABASHKA_PRELOADS_TEST=true
 echo "running tests part 2"
-call lein test :only babashka.main-test/preloads-test
+call lein test :only babashka.main-test/preloads-test || exit /B 1
 
 set BABASHKA_PRELOADS=(defn ithrow [] (/ 1 0))
 set BABASHKA_PRELOADS_TEST=true
 echo "running tests part 3"
-call lein test :only babashka.main-test/preloads-file-location-test
+call lein test :only babashka.main-test/preloads-file-location-test || exit /B 1
+
+set BABASHKA_PRELOADS=(require '[env-ns])
+set BABASHKA_CLASSPATH_TEST=true
+set BABASHKA_CLASSPATH=test-resources/babashka/src_for_classpath_test/env
+echo "running tests part 4"
+call lein test :only babashka.classpath-test/classpath-env-test || exit /B 1
+
+echo "not running pod tests on windows (yet)"
+REM set BABASHKA_POD_TEST=true
+REM call lein test :only babashka.pod-test || exit /B 1
+
+set BABASHKA_SOCKET_REPL_TEST=true
+call lein test :only babashka.impl.socket-repl-test || exit /B 1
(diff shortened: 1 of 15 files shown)


borkdude:
@bobisageek Excellent ideas.

The nrepl server test is failing it seems:

```
ERROR in (nrepl-server-test) (server.clj:11)
Uncaught exception, not in assertion.
expected: nil
  actual: java.lang.NullPointerException: null
 at babashka.nrepl.server$stop_server_BANG_.invokeStatic (server.clj:11)
    babashka.nrepl.server$stop_server_BANG_.invoke (server.clj:10)
    babashka.impl.nrepl_server_test$fn__36927.invokeStatic (nrepl_server_test.clj:209)
```

bobisageek:
I'm currently trying to repro the failure locally, but if that doesn't work, I might end up just having to skip that test on Windows for now. I was thinking maybe AppVeyor prevented opening ports, but there are other tests that passed that do start a server, so that doesn't seem like the culprit

borkdude:
That sounds good.

bobisageek:
For the time being, I'm going to skip the nrepl-server-test, and I'm curious to see if the socket-repl-test fails - there's nothing that comes across as obviously wrong to me, but apparently the server is nil when stop-server! is invoked.

bobisageek:
@borkdude Any thoughts on this new failure? I don't think anything actually changed in the native-image build, so not resolving a namespace seems unrelated to the tests... I'm going back over the commits to see if any of the changed files seem connected at all...

```
set BABASHKA_EDN=.build/bb.edn
bb release-artifact %zip%
Type:     java.lang.Exception
Message:  Could not find namespace: babashka.release-artifact.
Location: 22:56
```

borkdude:
@bobisageek Not sure what this is. Meanwhile another build ran successfully: https://ci.appveyor.com/project/borkdude/babashka/builds/39670942

It might just be a glitch. I'm going to trigger your build once again.

borkdude:
It's running here: https://ci.appveyor.com/project/borkdude/babashka/builds/39670985

script/test.bat
```diff
@@ -11,16 +11,32 @@ set PATH=%GRAALVM_HOME%\bin;%PATH%
...
 
 set BABASHKA_PRELOADS=(defn ithrow [] (/ 1 0))
 set BABASHKA_PRELOADS_TEST=true
 echo "running tests part 3"
-call lein test :only babashka.main-test/preloads-file-location-test
+call lein test :only babashka.main-test/preloads-file-location-test || exit /B 1
+
+set BABASHKA_PRELOADS=(require '[env-ns])
+set BABASHKA_CLASSPATH_TEST=true
+set BABASHKA_CLASSPATH=test-resources/babashka/src_for_classpath_test/env
```
borkdude, line 33: @bobisageek It might be failing because of this. I think you need to unset the classpath at the end of this file.

bobisageek: yeah - this I can repro locally, so I'll go through and unset those. I'm very sorry for cutting in to your Saturday like this

borkdude: @bobisageek No worries!

bobisageek: ok, we're past that hurdle - I think that the next failure is because BABASHKA_EDN is set when that test runs, so I'm testing that locally now, and I'll push a change to appveyor.yml if my test supports that theory

borkdude: Could very well be. In any case it's probably good to unset it after executing the github artifact script.

bobisageek: interestingly, unsetting BABASHKA_EDN fixes the test locally, but not on AppVeyor, so for now, I'm just skipping it, and maybe some hammock time will surface some new ideas to try

PR #902, opened 2021-06-24
add more windows testing (#235)

bobisageek:
- add default test selector to skip "windows only" tests
- in cases where the differences between *nix shell and windows shell
  make the test very messy, add a separate "windows only" test
- make more tests work on Windows

all my local CI testing passes, but I guess we'll see if AppVeyor agrees with that assessment


borkdude:
Good work!

Issue #369, opened 2020-04-20
Write some docs on using clojure.data.xml


borkdude:
Example:

``` clojure
$ bb '
(->> (curl/get "http://planet.clojure.in/atom.xml")
  :body (#(xml/parse-str % :skip-whitespace true))
  :content (filter #(= "entry" (some-> % :tag name)))
  first :content first :content)'

("http://PurelyFunctional.tv Newsletter 374: Seek truth, not convenience")
```

Issue #324, opened 2020-04-03
Document checking if data is available on stdin


borkdude:
```
$ bb '(pos? (.available System/in))'
false
$ echo true | bb '(pos? (.available System/in))'
true
```

cldwalker:
I noticed this detection doesn't work when chaining one bb command to another:

```sh
$ bb '(println [1 2 3])' | bb '(.available System/in)'
0
$ echo '[1 2 3]' | bb '(.available System/in)'
8
```
Any idea why this would happen?

kolharsam:
@cldwalker I tried the code snippet that you posted.
```clojure
$ bb '(println [1 2 3])' | bb '(.available System/in)'
=> 8
```
This seems to be giving the correct result on `v0.0.94`.

borkdude:
Maybe it depends on the operating system. What operating system are you using @kolharsam? I still see:

```
$ bb '(println [1 2 3])' | bb '(.available System/in)'
0
```

kolharsam:
I'm on macOS 10.15.4.


> Maybe it depends on the operating system.

I thought so too.

cldwalker:
I'm on osx 10.13.1. For now I worked around this by adding an explicit flag to signal to not read from stdin. Of course it would be nice if we could detect this between bb commands

borkdude:
@cldwalker Another thing I noticed:

```

$ bb '(println "foo")' | bb '(.available System/in)'
4
```
so when printing a string, it does seem to work.

borkdude:
This also seems to work:

```
$ bb '(println ["foo"])' | bb '(Thread/sleep 100) (.available System/in)'
6
```
so it must be a matter of timing.

PR #911, opened 2021-07-02
add examples for #324 and #369 [skip ci]

bobisageek:
- checking stdin for available input
- using data.xml lib

the XML example uses a contrived XML doc, but I thought that might be nice from a "fits in my head" perspective


borkdude:
Thanks, I will look at this hopefully soon. The examples have evolved since those issues were created, so we already have some XML examples (reading a version from a pom.xml file). The other example about `.available`, I'm not sure if we should recommend it, since the issue showed it's kind of brittle.

borkdude:
Merged it anyway. Thanks!

Issue #839, opened 2021-05-13
Add ffclj to list of library


borkdude:
https://github.com/luissantos/ffclj

Issue #806, opened 2021-04-27
Add logger example


borkdude:
https://gist.github.com/borkdude/c97da85da67c7bcc5671765aef5a89ad

PR #924, opened 2021-07-07
ffclj link and logger example (#839 and #806)

bobisageek:


borkdude:
@bobisageek Thanks! For only editing `.md` files, please add `[skip ci]` to the commit message, this will save some cycles on CircleCI servers. Keep up the good work!

Issue #922, opened 2021-07-07
formatter-out support for clojure.pprint


borkdude:
The fork of data.json here: https://github.com/babashka/data.json runs with babashka's `formatter-out` branch.
Clojure.data.json uses `formatter-out` for pprint-ing JSON.

But using `formatter-out` in babashka/SCI yields mangled output, why?

``` clojure
user=> (babashka.main/main "-e" "(require '[clojure.pprint :as pprint])  (def print-array (pprint/formatter-out \"~<[~;~@{~w~^, ~:_~}~;]~:>\")) (pprint/with-pprint-dispatch #(if (seqable? %) (print-array %) (print %)) (with-out-str (pprint/pprint [1 2 3])))")
"[1,23 , ]\n"
```

Equivalent to executing the following file:
``` clojure
(require '[clojure.pprint :as pprint])

(def print-array (pprint/formatter-out "~<[~;~@{~w~^, ~:_~}~;]~:>"))

(pprint/with-pprint-dispatch
  #(if (seqable? %)
     (print-array %)
     (print %))
  (pprint/pprint [1 2 3]))
```

bobisageek:
In case it helps, it seems like there might be a few slightly different behaviors here... I experimented with some longer seqs and different separator characters to look for patterns:

It seems like a general pattern is that we get the opening bracket, then the first item in the seq, then the separator*, then the rest of the seq, then all the remaining separators.

All these examples are pprinting `[1 2 3 4]`, with only the separator (between `~^` and `~:`) changing:

", " (same as original example):
`[1,234 , , ]` - some character(s) other than a space followed by a space seems to just get the separator up to the space after the first element

"<>"
`[1<>234<><>]` - separators that don't end in a space seem to get the whole separator after the first element

" ,"
`[1 ,234 , ,]` - a separator that starts with a space seems to follow the same pattern as arbitrary characters

" "
`[1234   ]` - one or more spaces as a separator seems to push all the separators after all the elements (similar behavior comes from using two or three spaces as a separator)

bobisageek:
This is sort of a stab in the dark, so feel free to just ignore it if it doesn't make sense... any possibility that pprint/write-option-table, used in new-write, referring to "proper" clojure's `*print-pprint-dispatch*` and `*print-right-margin*` vars instead of the sci ones is somehow involved here? I'm playing around a little bit trying to determine how to test that possibility.

borkdude:
[...]

PR #931, opened 2021-07-10
add cl-format tests for nil and Writer args

bobisageek:
following up on your point in #922 about adding some extra tests, this just exercises the other options for the first arg to `cl-format`


borkdude:
Thanks!

PR #933, opened 2021-07-11
run lib tests in Windows CI

bobisageek:
From my local testing, it seems like the environment setup should be ok, but we'll see what Appveyor makes of it. :)

The change in the selmer test is due to what feels like a very esoteric issue with the URL paths for resources - I figure the testing is nicer to have, and then I'll look deeper into the issue once the tests are setup.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -71,6 +71,8 @@ build_script:
 
     call script/test.bat
 
+    call script/run_lib_tests.bat
+
 artifacts:
 - path: babashka-*-windows-amd64.zip
   name: babashka
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,6 +17,7 @@ cache:
   - 'graalvm -> appveyor.yml'
 
 clone_script:
+- cmd: git config --global core.autocrlf true
 - ps: >-
     if(-not $env:APPVEYOR_PULL_REQUEST_NUMBER) {
       git clone -q --branch=$env:APPVEYOR_REPO_BRANCH https://github.com/$env:APPVEYOR_REPO_NAME.git $env:APPVEYOR_BUILD_FOLDER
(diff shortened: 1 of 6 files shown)


borkdude:
@bobisageek It looks like most of them are passing :) Except for the \r\n thing, so we probably need a normalize-like feature in those lib tests.

bobisageek:
Yeah - there is a 'fix line separator' function in there, but I'm guessing that the git checkout isn't using the system-dependent line separator, because the 'expected' has the windows-specific line endings, but the 'actual' doesn't, which is created from rendering the file. I guess it's sort of "six in one, half-dozen in the other", but should we setup "git config --global core.autocrlf true" so that Appveyor checks out with Windows line endings, which would probably be more representative of what might happen in the course of using `bb` on a Windows system?

borkdude:
@bobisageek Why didn't we have these problems with the other tests?

bobisageek:
In the general case, a lot of the bb tests aren't "raw" rendering files the way selmer does - the 'base' tests that run `bb` would _typically_ be using clojure print functions, which would put the system-dependent line endings into the output, but with selmer, we're often templating the exact contents of a file, which preserves whatever line ending that file happened to have.

As an example, when I run the lib tests in WSL, those same tests fail in the other direction, because my WSL console is still reading the file from the same checkout I have in Windows, which has CRLF line endings, but the `fix-line-sep` function expects whatever System/lineSeparator is. So `fix-line-sep` says that the line ending should be \n (because it think it's on linux), but the file has \r\n because it's checked out on Windows (with autocrlf setup in git).

borkdude:
That makes sense.

bobisageek:
Now I just have to hope that this build works so that I didn't write up that whole explanation only to be wrong. :)

borkdude:
Yay! `{:test 845, :pass 5031, :fail 0, :error 0}`

Issue #936, opened 2021-07-13
Fix Windows native test when passing :env in clojure


borkdude:
https://github.com/babashka/babashka/blob/24d1de51a018fc0f3576febe6ae0eee76f6619d4/test/babashka/deps_test.clj#L83

/cc @bobisageek

bobisageek:
I think that at least a contributing factor here might be `process` handling backslashes in a somewhat awkward way.

In order to understand it better, I tried to peel away all the abstractions and dependencies, and I've got it down to a _relatively_ simple repro, based on the test, that can be run in bb:

```clj
(require '[babashka.process :as p])
(require '[babashka.fs :as fs])
(require '[clojure.java.shell :as shell])

(let [libs-dir (fs/file (fs/create-temp-dir) ".gitlibs")
      template (pr-str '(do (babashka.deps/clojure 
                             ["-Sforce" "-Spath" "-Sdeps" "{:deps {babashka/process {:git/url \"https://github.com/babashka/process\" :sha \"4c6699d06b49773d3e5c5b4c11d3334fb78cc996\"}}}"]
                             {:out     :string
                              :env-key {"PATH"    (System/getenv "PATH")
                                        "GITLIBS" :gitlibs}}) nil))
      a (-> template (str/replace ":gitlibs" (pr-str (str libs-dir)))
            (str/replace ":env-key" ":env"))
      _ (println a)]
  (println "===Process run===")
  (println @(p/process ["bb" a] {:out :string :err :string}))
  (println "===Sh run===")
  (println (shell/sh "bb" (pr-str a))))
```

On Windows, the Process run fails with that same `Unsupported escape character: \h` message. But with the repro, I can also see the `:cmd` from the process object:
`:cmd [bb (do (babashka.deps/clojure [\"-Sforce\" \"-Spath\" \"-Sdeps\" \"{:deps {babashka/process {:git/url \\"https://github.com/babashka/process\\" :sha \\"4c6699d06b49773d3e5c5b4c11d3334fb78cc996\\"}}}\"] {:out :string, :env {\"PATH\" (System/getenv \"PATH\")...`

If you look at the URL and sha in the :deps map, it looks like the backslashes are wrong... there should be three backslashes before the double-quotes, correct?

I'm going to keep playing and try to get down to a really small repro, but maybe this will provide some ideas on which to start building.

[...]

PR #950, opened 2021-07-21
add windows-specific babashka.deps-test/clojure-test (#936)

bobisageek:
This should address the skipped test in #936 - it's definitely not as pretty as using `pr-str`, but I haven't been able to get a `pr-str` working on Windows (native) because of the differences between Windows shell escaping and Clojure escaping. I will not be offended if you want to reject the code based purely on the relative ugliness of the string manipulation. :)


borkdude:
Thanks!

Issue #940, opened 2021-07-15
Windows tests: disabled lib tests


borkdude:
FYI @bobisageek, I kept getting this problem in CI:

<img width="972" alt="Screenshot 2021-07-15 at 16 40 00" src="https://user-images.githubusercontent.com/284934/125807110-0198baa5-8778-400f-be8f-e0e797ca4df7.png">

I didn't investigate yet why this happened, but I decided to turn off the lib tests.

I also disabled the JVM tests for now on Windows to make the build slightly faster: if the native tests pass I think that's sufficient.

bobisageek:
I'm able to re-create this, but only on newer commits (based on a bisect). The symptom I'm seeing locally is that for a reason (that I don't currently know), the output of `bb clojure...` doesn't get put into the variable, so most of the classpath is missing.

I'm currently thinking about how to minimize the reliance on Windows batch behavior, in large part because I can't repro the behavior on the same revs where appveyor exhibits it.

borkdude:
I probably broke it myself in the process of getting 0.5.0 out the door, sorry... :)
I'm looking forward to your thoughts of relying less on Windows .bat. Perhaps we could just make a script in Clojure and run it on the JVM, although that will be a tad slower.

borkdude:
Or just use the compiled bb for scripting, I would be fine with that. If the native tests succeed on Windows, I'm convinced that the JVM tests would also have succeeded, so we can just start from the native.

bobisageek:
To avoid the "for loop" thing, I put together a little bb script that does a stripped-down version of what the clojure command does - read deps.edn and merge the deps/paths with lib-tests alias deps and paths, and then I'm just spitting that out to a file to be used as a bb.edn.

Just a quick question - I'm a bit unclear on why `test-resources/lib_tests` is added to the classpath in the shell scripts themselves - I'm wondering if it would make sense to add that path to `:extra-paths` in deps.edn instead, and then both of the lib-test scripts wouldn't need that manually added to the classpath. I can make it work either way, but if you agree that moving the path to deps.edn is cleaner, then those scripts get just a little bit nicer IMO.

borkdude:
> Just a quick question - I'm a bit unclear on why test-resources/lib_tests is added to the classpath in the shell scripts themselves 

Can you show me where this happens?

bobisageek:
[...]

PR #956, opened 2021-07-25
fix windows lib tests (#940)

bobisageek:
- add lib-tests path to deps.edn and remove it from shell scripts
- change windows lib test batch script to write a bb.edn file
- re-enable native lib tests on windows

things are working locally, and the main reliance on the windows stuff is environment variables, and not capturing command output

in my local testing, the (native) lib tests run in about 30 seconds, so I don't imagine these will add a ton of time to the build

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -74,7 +74,7 @@ build_script:
 
     call script/test.bat
 
-# call script/run_lib_tests.bat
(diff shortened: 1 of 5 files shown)


bobisageek:
If you're good with these changes, I think #940 and #936 can be closed

borkdude:
Thanks a lot!

PR #964, opened 2021-08-03
add META-INF resource to windows compilation (fixes #963)

### LINKED ISSUE #963: bb-edn-test/print-deps-test fails on Windows
In case you want to look at this, /cc @bobisageek.

bobisageek:
It looks like the deps.edn resource that got added to `compile` just needs to get added to `compile.bat` as well. This made the test pass immediately in my local env.


borkdude:
Good catch. I guess we'll have to make a new bb release for the print-deps stuff to work on Windows.

Issue #957, opened 2021-07-26
Some vars are missing ns metadata.


borkdude:
By @cursive-ide on Slack:

However, in bb for a bunch of vars `(#'global-hierarchy #'*ns* #'*print-meta* #'*out* #'alength #'*math-context* #'*command-line-args* #'*err* #'*print-level* #'*in* #'*print-length* #'*print-namespace-maps* #'*warn-on-reflection* #'*data-readers* #'*file* #'clojure.core/unquote)` I get:
```
user=> (meta #'*ns*)
{:dynamic true}
```

Find them all:

```
user=> (clojure.pprint/pprint (for [ns (all-ns), [k v] (ns-publics ns) :when (not (:ns (meta v)))] [(str ns) k]))
(["babashka.signal" pipe-signal-received?]
 ["clojure.java.shell" with-sh-dir]
 ["clojure.java.shell" with-sh-env]
 ["user" *input*]
 ["babashka.wait" wait-for-port]
 ["babashka.wait" wait-for-path]
 ["clojure.core.server" repl]
 ["clojure.core.server" prepl]
 ["clojure.core.server" io-prepl]
 ["clojure.core.server" start-server]
 ["clojure.stacktrace" root-cause]
 ["clojure.stacktrace" print-trace-element]
 ["clojure.stacktrace" print-throwable]
 ["clojure.stacktrace" print-stack-trace]
 ["clojure.stacktrace" print-cause-trace]
 ["clojure.test" test-all-vars]
 ["clojure.test" do-report]
 ["clojure.test" run-all-tests]
 ["clojure.test" assert-any]
 ["clojure.test" testing-contexts-str]
 ["clojure.test" join-fixtures]
 ["clojure.test" assert-expr]
 ["clojure.test" compose-fixtures]
 ["clojure.test" assert-predicate]
 ["clojure.test" function?]
 ["clojure.test" test-vars]
 ["clojure.test" successful?]
 ["clojure.test" use-fixtures]
 ["clojure.test" inc-report-counter]
 ["clojure.test" testing-vars-str]
 ["clojure.test" test-ns]
 ["clojure.test" run-tests]
 ["clojure.core.protocols" Datafiable]
 ["clojure.core.protocols" Navigable]
 ["clojure.java.browse" *open-url-script*]
 ["clojure.java.browse" browse-url]
 ["clojure.data.csv" read-csv]
 ["clojure.data.csv" write-csv]
 ["clojure.core" macroexpand]
[...]

PR #967, opened 2021-08-08
test that var metadata matches containing ns

bobisageek:
- `quick-check`'s metadata had its ns as `clojure.test.check.properties`, but is in `clojure.test.check`, so change that meta
- add `namespace-test` to test that all the public vars have `:ns` metadata, and that it matches their containing namespace (checking that ns names are equal, not that they're identical, because the features don't currently have identical namespaces)

I would say, at your discretion, this might be sufficient to close out #957 - the only vars that don't have `:ns` in their meta now are the `clojure.lang` ones and `*input*`

### DIFF
--- a/feature-jdbc/babashka/impl/jdbc.clj
+++ b/feature-jdbc/babashka/impl/jdbc.clj
@@ -2,7 +2,7 @@
   {:no-doc true}
   (:require [next.jdbc :as njdbc]
             [next.jdbc.sql :as sql]
-            [sci.impl.namespaces :refer [copy-var]]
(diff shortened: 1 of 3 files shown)


bobisageek:
I missed a var on `next.jdbc` - working on that now

bobisageek:
My local test scripts were missing the feature flags used by circleCI, so that's why I missed the macro on next.jdbc. There is also a small issue the lanterna ns, but that's not tested by default, so I'll just include that fix in a future PR with docs or something.

Issue #866, opened 2021-05-25
Bump image used to build Native Image after Debian 11 is released


thiagokokada:
**Is your feature request related to a problem? Please describe.**
Nowadays we use old stable version of Debian (`stretch`, or Debian 9). There will be no more support after 30 June 2022, so it makes sense that we should update it eventually.

It doesn't make much sense to update it right now to Debian 10 since this will not bring much benefits (the most impactful thing would be the upgrade from GCC 6.3 to GCC 8.4). However with the release of Debian 11 we will have two things that we end up using:

- `gcc` will be upgraded to 10.2
- `musl-tools` will be upgraded to 1.22

Today we backport `musl-tools` from unstable, but since the version from unstable and the version from Debian 11 are the same (1.22), it means that after Debian 11 we can drop [this hack](https://github.com/babashka/babashka/blob/master/script/setup-musl#L15-L36) and just install `musl-tools` directly from repositories.

So I think upgrading Debian 11 after its release will be worth it.

**Describe the solution you'd like**
Change the image from [`Dockerfile`](https://github.com/babashka/babashka/blob/master/Dockerfile#L1) to whatever version is released on Clojure official images after Debian 11 is released. Also need to update the image used on CircleCI.

This probably also means we need to upgrade `leiningen` to a newer version (since newer images are generally only build with the latest `leiningen` available). The current version is 2.9.6, but this may change after Debian 11 is released.

**Describe alternatives you've considered**
We can continue to build binaries using Debian 9, however eventually we will need to update the image after Debian 9 stops receiving updates.

**Additional context**
N/A

thiagokokada:
This issue can probably have a `blocker` label or something, since there is still some months until Debian 11 is finally released.

thiagokokada:
One thing that I remember is that bumping Debian will also bump the `glibc` version (Debian 9 uses 2.24, Debian 11 uses 2.31). This shouldn't be too much a problem since 2.24 and 2.31 have the same minimum kernel version (3.2), and AFAIK Debian compiles `glibc` explicitly with this minimum version.

But something to keep in mind to test after switching.

thiagokokada:
CC @lispyclouds.

ieugen:
Debian 11 release in 1 week.

thiagokokada:
Opened an issue on `docker-clojure` repository for Debian 11 (bullseye) is used on the image instead of Debian 10 (stretch): https://github.com/babashka/babashka/issues/866

[...]

PR #972, opened 2021-08-22
make the REPL requires available to command-line scripts (closes #729)

### LINKED ISSUE #729: Add REPL requires to user namespace for one-liners
Map the REPL requires `(pprint, etc,)` also to the user namespace, so you can just write `(pprint *input*)`.

bobisageek:
I assume this is what you had in mind for #729

### DIFF
--- a/src/babashka/main.clj
+++ b/src/babashka/main.clj
@@ -792,6 +792,9 @@ Use bb run --help to show this help output.
                        expressions
                        (sci/binding [sci/file abs-path]
                          (try
+                           ; when evaluating expression(s), add in repl-requires so things like
+                           ; pprint and dir are available
+                           (sci/eval-form sci-ctx `(apply require (quote ~clojure-main/repl-requires)))
                            (loop []
                              (let [in (read-next *in*)]
                                (if (identical? ::EOF in)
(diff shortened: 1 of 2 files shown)


bobisageek:
It looks like musl got updated in the last day or so (failing build uses 1.2.2-4, last build on master used 1.2.2-1)... I spent some time trying to figure out what changes are needed, but I'm not having a lot of luck. Given the nature of these changes, I'm imagining that the musl build (with the new version) will probably fail on master as well.

I'll see if I have any more luck with musl after I get some sleep, but it's sort of a blind spot for me, so I'm very open to suggestions. 😄

borkdude:
@bobisageek Hm yeah, same problem with clj-kondo: https://github.com/clj-kondo/clj-kondo/issues/1365

lispyclouds:
Maybe its something to do with the recent Debian 11 release. I will have a look today.

thiagokokada:
Probably now that Debian 11 is released, unstable started to finally do big changes.

My suggestion is either to backport musl from Debian 11 release (should be the correct version), or better yet, just migrate to Debian 11 instead of using Debian 9: https://github.com/babashka/babashka/issues/866.

thiagokokada:
> Probably now that Debian 11 is released, unstable started to finally do big changes.
> 
> My suggestion is either to backport musl from Debian 11 release (should be the correct version), or better yet, just migrate to Debian 11 instead of using Debian 9: #866.

Sadly, Clojure's docker image is still being compiled with Buster: https://hub.docker.com/_/clojure/

bobisageek:
pulled in the changes from master - I was able to do a musl build with those changes on debian in WSL, so I think this should work now

borkdude:
Thanks!

PR #974, opened 2021-08-22
script: use musl from bullseye instead of unstable

thiagokokada:
Now that bullseye was released, Debian is starting to make bigger changes on unstable. To avoid breakage, let's use musl from bullseye that is the correct version that we were using until now (1.22-1): https://packages.debian.org/bullseye/musl-tools

Eventually a better solution would be to simply use Debian bullseye to compile everything, however the official Clojure image from Docker is still using stretch. So this is a temporary solution until them.

Fix issue: https://github.com/babashka/babashka/pull/972

### DIFF
--- a/script/setup-musl
+++ b/script/setup-musl
@@ -12,13 +12,14 @@ if [[ -z "${BABASHKA_MUSL:-}" ]]; then
     exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


borkdude:
Thanks a lot!

PR #983, opened 2021-08-28
Jdk 11 http client

grzm:


grzm:
I'm not sure what's going on with `linux-aarch64-static`: I don't have a CirceCI account so I can't see the details. Given the others are passing, I suspect it's something more than what I'd know how to deal with anyway.

borkdude:
@grzm I recommend making an account so you can see what's going on in the builds.

Issue #981, opened 2021-08-28
clojure.tools.logging.readable


corasaurus-hex:
It would be nice to have [clojure.tools.logging.readable](https://github.com/clojure/tools.logging/blob/master/src/main/clojure/clojure/tools/logging/readable.clj) included with the bundled clojure.tools.logging. It's just for convenience, but it is _so_ convenient. This namespace provides logging where if you do `(log/info & args)` and any of the `args` aren't strings then each non-string `arg` is run through `pr-str` before being concatenated with the rest of the message. This means you can log most anything without having to wrap it in a `(pr-str)` yourself manually.

borkdude:
Fixed by @bobisageek, thanks!

PR #987, opened 2021-08-29
Logging feature flag and tools.logging.readable (#981)

bobisageek:
I'd say please double-check me on the migration to the feature flag - I've looked over it about 10 times myself, but if I missed it once, there's a good chance I missed it nine more times. :)

Regarding the tweaks to the test programs in the `logging-test` namespace: the reason I made those was to essentially reset the logger state between the test programs. When I added the new test, it caused some of the old tests to fail on JVM, I believe because the log level/config persisted across tests.


test/babashka/logging_test.clj
```diff
@@ -38,34 +40,77 @@
...
+  (println "with print-readably set to nil (overridden by log macros)")
+  (binding [*print-readably* nil]
+    (test-fn))
+
+  (println "setting log level")
+  (timbre/set-level! :warn)
+  (test-fn)
+  (timbre/set-level! :debug)))
+
+(when features/logging?
```
borkdude, line 106: @bobisageek To be sure, I would leave this condition out, as the tests always run with the default feature flags and if we accidentally break the feature flag, the tests won't run and that will go undetected.

borkdude: At least, I don't think we have these conditions for other default-enabled features so it would be more in line with that, but maybe we should. Right now we don't at least.

bobisageek: Sounds good. I'm re-running my local test battery, so hopefully I'll have a new commit shortly.

PR #1005, opened 2021-09-09
Use Debian bullseye on CI/CD/Dockerfile

thiagokokada:
Fix issue #866.

Also, many QoL changes on `Dockerfile`/`.cicleci/config.yml`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ jobs:
     working_directory: ~/repo
     environment:
       LEIN_ROOT: "true"
-      GRAALVM_HOME: /home/circleci/graalvm-ce-java11-21.2.0
+      GRAALVM_HOME: /root/repo/graalvm-ce-java11-21.2.0
       BABASHKA_PLATFORM: linux # used in release script
       BABASHKA_TEST_ENV: native
       BABASHKA_XMX: "-J-Xmx6500m"
@@ -127,7 +127,7 @@ jobs:
     working_directory: ~/repo
     environment:
       LEIN_ROOT: "true"
-      GRAALVM_HOME: /home/circleci/graalvm-ce-java11-21.2.0
+      GRAALVM_HOME: /root/repo/graalvm-ce-java11-21.2.0
       BABASHKA_PLATFORM: linux # used in release script
       BABASHKA_TEST_ENV: native
       BABASHKA_STATIC: "true"
@@ -203,7 +203,7 @@ jobs:
     working_directory: ~/repo
     environment:
       LEIN_ROOT: "true"
-      GRAALVM_HOME: /home/circleci/graalvm-ce-java11-21.2.0
+      GRAALVM_HOME: /root/repo/graalvm-ce-java11-21.2.0
       BABASHKA_PLATFORM: linux # used in release script
       BABASHKA_ARCH: aarch64
       BABASHKA_TEST_ENV: native
@@ -274,7 +274,7 @@ jobs:
     working_directory: ~/repo
     environment:
       LEIN_ROOT: "true"
-      GRAALVM_HOME: /home/circleci/graalvm-ce-java11-21.2.0
+      GRAALVM_HOME: /root/repo/graalvm-ce-java11-21.2.0
       BABASHKA_PLATFORM: linux # used in release script
       BABASHKA_ARCH: aarch64
       BABASHKA_TEST_ENV: native
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -217,8 +217,6 @@ jobs:
     working_directory: ~/repo
     environment:
       LEIN_ROOT: "true"
-      GRAALVM_VERSION: "21.2.0"
-      GRAALVM_HOME: /home/circleci/graalvm-ce-java11-21.2.0
       BABASHKA_PLATFORM: linux # used in release script
       BABASHKA_ARCH: aarch64
       BABASHKA_TEST_ENV: native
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   jvm:
     docker:
       # specify the version you desire here
-      - image: circleci/clojure:lein-2.9.1
+      - image: clojure:openjdk-11-lein-2.9.6-bullseye
```
thiagokokada, line 10: CircleCI's Clojure image is out-of-date (it doesn't support `bullseye`), and also seems to be considered legacy: https://circleci.com/docs/2.0/circleci-images/#legacy-language-images

.circleci/config.yml
```diff
@@ -30,11 +30,11 @@ jobs:
           command: |
             wget -nc https://download.clojure.org/install/linux-install-1.10.1.447.sh
             chmod +x linux-install-1.10.1.447.sh
-            sudo ./linux-install-1.10.1.447.sh
```
thiagokokada, line 33: `sudo` shouldn't be necessary (since all commands should run as root already), and also it isn't included on the `clojure:openjdk-11-lein-2.9.6-bullseye` image.

.circleci/config.yml
```diff
@@ -96,16 +96,16 @@ jobs:
...
           command: |
-            sudo apt-get install lsof
+            apt-get install lsof
       - run:
           name: Install native dev tools
           command: |
-            sudo apt-get update
-            sudo apt-get -y install g++ zlib1g-dev
+            apt-get update
+            apt-get -y install build-essential zlib1g-dev
```
thiagokokada, line 108: `clojure:openjdk-11-lein-2.9.6-bullseye` lacks some tooling (like `make`), and installing `build-essential` fill this gap. 

This also make CircleCI's config it closer to the `Dockerfile`.

.dockerignore
```diff
@@ -12,7 +12,6 @@ test/
 .gitmodules
 appveyor.yml
 CHANGES.md
-deps.edn
```
thiagokokada, line 15: I don't know how this worked before, but this file is necessary to build using the `Dockerfile`.

Dockerfile
```diff
@@ -1,16 +1,18 @@
-FROM clojure:lein-2.9.1 AS BASE
+FROM clojure:openjdk-11-lein-2.9.6-bullseye AS BASE
 
 ENV DEBIAN_FRONTEND=noninteractive
 RUN apt update
-RUN apt install --no-install-recommends -yy curl unzip build-essential zlib1g-dev sudo
```
thiagokokada, line 5: `curl`, `unzip` are already included on `clojure:openjdk-11-lein-2.9.6-bullseye`. `sudo` is not used anywhere in the `Dockerfile` (and also, all commands already run as root).

script/setup-musl
```diff
@@ -12,29 +12,7 @@ if [[ -z "${BABASHKA_MUSL:-}" ]]; then
...
-Pin-Priority: 700
-Package: *
-Pin: release a=testing
-Pin-Priority: 650
-Package: *
-pin: release a=bullseye
-pin-priority: 600
-eof
-
-apt-get update -y && apt-get install musl-tools/bullseye -y
```
thiagokokada, line 37: Not necessary to backport anymore since we are already using bullseye.

.circleci/config.yml
```diff
@@ -183,17 +175,13 @@ jobs:
           command: |
             wget https://download.clojure.org/install/linux-install-1.10.1.447.sh
             chmod +x linux-install-1.10.1.447.sh
-            sudo ./linux-install-1.10.1.447.sh
-      - run:
-          name: Install lsof
-          command: |
-            sudo apt-get install lsof
```
thiagokokada, line 190: Not used anywhere and not available on `clojure:openjdk-11-lein-2.9.6-bullseye` by default.

.circleci/config.yml
```diff
@@ -354,26 +310,12 @@ jobs:
...
-      - run:
-          name: Install Leiningen
-          command: |
-            sudo script/install-leiningen
-      - run:
-          name: Install Clojure
-          command: |
-            wget https://download.clojure.org/install/linux-install-1.10.1.447.sh
-            chmod +x linux-install-1.10.1.447.sh
-            sudo ./linux-install-1.10.1.447.sh
```
thiagokokada, line 366: Already included on the base image.

thiagokokada:
Huh... I thought that CircleCI built everything as root, but it seems it builds with a non-root user. So this approach will not work, sadly.

borkdude:
@thiagokokada You can use `sudo` though.

thiagokokada:
> @thiagokokada You can use `sudo` though.

It is not available on `clojure:openjdk-11-lein-2.9.6-bullseye` by default, and we have a chicken-egg problem for installing it (I can't use `sudo apt install sudo` without `sudo`).

thiagokokada:
Actually, it does seem that the commands without `sudo` worked, so the problem seems something else. Needs more investigation...

.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   jvm:
     docker:
       # specify the version you desire here
-      - image: circleci/clojure:lein-2.9.1
+      - image: circleci/clojure:lein-2.9.6
```
thiagokokada, line 10: `jvm` needs PostgreSQL to run, so using the `circleci/clojure:lein-2.9.6` that already includes it by default.

thiagokokada:
@borkdude Well, it is working now (sorry for all the runs on CircleCI). BTW, what I said above was wrong, the error was purely my fault.

Not entirely happy with this PR, will wait some days to see if CircleCI releases their updated image with more tooling included. However if they do not, I will open the PR for review later.

borkdude:
Thanks so far, I'll wait.

borkdude:
@thiagokokada Could you perhaps not use force push? I don't think it adds anything, but it makes following what's going on a lot harder for me. The Github UI shows "this isn't here anymore"-like things when I click on the notifications.

thiagokokada:
Sorry @borkdude, will avoid this from now on.

thiagokokada:
I think for clean-ups is pretty much it, I will just wait to see if CircleCI will release a new version of their Clojure image.

thiagokokada:
@borkdude It is ready now.

borkdude:
@thiagokokada Beautiful, thank you!

PR #1012, opened 2021-09-19
XML edn reading and populating *data-readers* for feature deps

### LINKED ISSUE #1004: roundrip from xml to edn and back is broken
**version**
0.6.0

**platform**
Linux

**problem**
`*data-readers*` not set for xml package, similar to #583 

**repro**

    bb '(print "<a><b>data</b><b>data1</b></a>")' | bb '(xml/parse *in*)' | bb '(read *input*)'

&#x2013;&#x2014; Error ---------------------------------------------------------------&#x2013;&#x2014;
Type:     clojure.lang.EdnReader$ReaderException
Message:  java.lang.RuntimeException: No reader function for tag xml/element
Location: 1:1

    echo "<a><b>data</b><b>data1</b></a>" | bb '(xml/parse *in*)' | bb '(read *in*)'

&#x2013;&#x2014; Error ---------------------------------------------------------------&#x2013;&#x2014;
Type:     java.lang.Exception
Message:  No reader function for tag xml/element
Location: 1:1

**expected behavior**

    echo "<a><b>data</b><b>data1</b></a>" | bb '(xml/parse *in*)' | clj -Sdeps '{:deps {org.clojure/data.xml {:mvn/version "0.2.0-alpha6"}}}' -e "(require 'clojure.data.xml) (read *in*)"

\#xml/element{:tag :a, :content [#xml/element{:tag :b, :content ["data"]} #xml/element{:tag :b, :content ["data1"]}]}

    clj -Sdeps '{:deps {org.clojure/data.xml {:mvn/version "0.2.0-alpha6"}}}' -e "*data-readers*"

\#:xml{ns #'clojure.data.xml.name/uri-symbol, element #'clojure.data.xml.node/tagged-element}

bobisageek:
Should close #1004 .

- On Clojure-JVM, `clojure.core/*data-readers*` has an initial value of an empty map. This change facilitates using `into` later on when we populate `*data-readers*` for 'runtime'.
- conditionally add the `clojure.data.xml` data readers to edn-readers
- the value of edn-readers mostly reflects what `*data-readers*` would be when starting Clojure-JVM with the YAML and XML deps, so we dump edn-readers into `*data-readers*` to emulate the handling of data_readers.clj (it's missing `ordered/set`, but I don't think that's needed for the YAML feature).
- add XML and YAML round-trip testing, both between bb invocations (to emulate piping and exercise edn reading to `*input*`) and within one bb process (to exercise `*data-readers*` usage)

### DIFF
--- a/src/babashka/impl/clojure/core.clj
+++ b/src/babashka/impl/clojure/core.clj
@@ -26,7 +26,7 @@
   ([sym] (core-dynamic-var sym nil))
   ([sym init-val] (sci/new-dynamic-var sym init-val {:ns clojure-core-ns})))
 
-(def data-readers (core-dynamic-var '*data-readers*))
+(def data-readers (core-dynamic-var '*data-readers* {}))
 (def command-line-args (core-dynamic-var '*command-line-args*))
 (def warn-on-reflection (core-dynamic-var '*warn-on-reflection* false))
 (def math-context (core-dynamic-var '*math-context*))
--- a/src/babashka/main.clj
+++ b/src/babashka/main.clj
@@ -430,6 +430,9 @@ Use bb run --help to show this help output.
                    (assoc 'xml/ns @(resolve 'clojure.data.xml.name/uri-symbol)
                           'xml/element @(resolve 'clojure.data.xml.node/tagged-element))))
 
+;; also put the edn readers into *data-readers*
+(sci/alter-var-root core/data-readers into edn-readers)
+
 (defn edn-seq*
   [^java.io.BufferedReader rdr]
   (let [edn-val (edn/read {:eof ::EOF :readers edn-readers :default tagged-literal} rdr)]
@@ -661,7 +664,7 @@ Use bb run --help to show this help output.
 (defn exec [cli-opts]
   (binding [*unrestricted* true]
     (sci/binding [core/warn-on-reflection @core/warn-on-reflection
-                  core/data-readers (into @core/data-readers edn-readers)
+                  core/data-readers @core/data-readers
                   sci/ns @sci/ns]
       (let [{version-opt :version
              :keys [:shell-in :edn-in :shell-out :edn-out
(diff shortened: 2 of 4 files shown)


bobisageek:
@borkdude sorry for the ping; I would like your feedback on two points before marking this ready:
- I realize that it feels a little weird at the moment to initialize `*data-readers*` to an empty map, just to pour the conditional map into it later - the reason I left it like that was because I figured if at some point there are unconditional data readers, then the initial map could have some entries, and edn-readers could add the conditional ones in, but it would likely be that tiny bit faster to just bind `edn-readers` with the trade-off of losing that little bit of flexibility. Do you have a preference in either direction there?
- A thing I thought about doing was putting the XML and YAML tests in the feature directories (e.g. making 'src' and 'test' directories under the feature folders to exclude those tests on the off chance that tests were ever run with those features flags disabled). However, this would set a bit of a precedent, so I wanted to get your thoughts first on if it's worth the changes to handle that relatively unlikely situation.

borkdude:
@bobisageek Thanks for the PR! I think the `into` at runtime could be become a `sci/alter-var-root` at the top level so we don't pay the execution time penalty (although it's probably minor: if you add up things over years and years of development, it might becomes significant).

About the feature flags: I'm fine with doing that, but perhaps do this in a separate PR so we keep things consistent.

bobisageek:
that's an excellent answer that didn't occur to me for some reason - thank you

For the feature test thing, I was also thinking a separate, focused PR might be better - also means I can go back and look for any other tests that should fall into that pattern (the logging feature comes to mind). 
I'm also thinking that, while I'm sort of tweaking feature flag-related stuff, if you're ok with it, I'd like to "clean up" some features that I don't think are currently candidates for being turned off. The two that I can recall off-hand are core.async and rewrite-clj - since they're used by tasks, I think I'm right in saying that bb won't compile with these features disabled.

borkdude:
Correct! Btw, is this ready for merge?

bobisageek:
yeah - assuming the windows tests don't explode for some unforeseen reason, it's good to go

PR #1014, opened 2021-09-23
make core.async and rewrite-clj integral

bobisageek:
With the addition of tasks, core.async and rewrite-clj became non-optional dependencies. This change "integrates" these two libs by removing their feature flags and bringing the code into the "main" package structure.


borkdude:
Thank you!

Issue #451, opened 2020-05-26
Document gzip usage


borkdude:
``` clojure
$ bb '(with-open [os (java.util.zip.GZIPOutputStream. (io/output-stream "/tmp/foo.zip"))] (.write os (.getBytes "dude")) (.finish os))'
$ bb '(slurp (java.util.zip.GZIPInputStream. (io/input-stream "/tmp/foo.zip")))'
"dude"
```

bobisageek:
Can we close this? I thought the mention on the PR would do it, but it looks like using `and` as a join only works on the other big git site.

PR #1016, opened 2021-09-26
Component lib test

### LINKED ISSUE #742: Test and document stuartsierra/component
Add component tests to lib tests (in so far they are compatible).

It looks like it works:

``` clojure 
(ns component
  (:require [babashka.deps :as deps]))

(deps/add-deps '{:deps {com.stuartsierra/component {:mvn/version "1.0.0"}}})

(require '[com.stuartsierra.component :as component])

(defrecord Component [x]
  component/Lifecycle
  (start [component]
    (prn :start)
    (assoc component :started true))
  (stop [component]
    (prn :stop)
    (dissoc component :started)))

(def sm (component/system-map :c1 (->Component :foo)))

(component/start sm)
;;=> :start
;;=> {:c1 {:x :foo, :started true}}

(component/stop sm)
;;=> :stop
;;=> {:c1 {:x :foo}}
```

bobisageek:
Closes #742 and #451 

- add a lib test for component
- add component to the libraries list
- add an example of using gzip streams to the examples page

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -89,7 +89,8 @@
                          org.clojure/data.json {:mvn/version "2.4.0"}
                          clj-commons/multigrep {:mvn/version "0.5.0"}
                          amperity/vault-clj {:mvn/version "1.0.4"}
-                         java-http-clj/java-http-clj {:mvn/version "0.4.3"}}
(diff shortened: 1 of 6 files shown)


borkdude:
Thanks!

PR #1021, opened 2021-09-29
print locals only when debug option is enabled (#1015)

### LINKED ISSUE #1015: Consider only printing locals when --debug is provided
Sometimes it's awkward when secrets are printed in error output, in locals.
We can switch this off by default and only do this when `--debug` is provided.
Similar for data from ex-data fields.

/cc @mk

bobisageek:
Closes #1015 

The issue title says "consider", so I don't know if you definitely want this.

### DIFF
--- a/src/babashka/impl/error_handler.clj
+++ b/src/babashka/impl/error_handler.clj
@@ -119,7 +119,7 @@
             (ruler "Context")
             (println ec)
             (println))
-          (when-let [locals (not-empty (:locals d))]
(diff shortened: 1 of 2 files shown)


bobisageek:
I don't _think_ the failure is related to these changes - seems like just a random-ish timeout - is there a way to maybe retry that build at some point?

borkdude:
@bobisageek Thanks! Also I think we should only print the ex-data when debug is enabled.

PR #1022, opened 2021-09-29
print ex-data map from error handler only when in debug mode

bobisageek:
follow-up from #1021 :
My apologies - I did read that note in the original #1015 description, but for whatever reason, I parsed the bit about printing exception data as "similar to the way bb currently only prints the stack trace and what not (the 'Exception' section of the printed output) when in debug mode", as opposed to what you meant (also suppressing the `ex-data` map if we're not in debug).


bobisageek:
sounds good

Issue #656, opened 2020-11-20
babashka.crypto / hmac


borkdude:
Currently you can create hmac with openssl. But maybe adding java crypto for this might be better.

``` clojure
(defn sign-with-string [key msg]
  (-> (sh "openssl" "dgst" "-sha256" "-mac" "hmac" "-macopt" (str "key:" key) :in msg)
      :out
      str/trim))
(defn sign-with-hex [key msg]
    (-> (sh "openssl" "dgst" "-sha256" "-mac" "hmac" "-macopt" (str "hexkey:" key) :in msg)
        :out
        str/trim))
```

You can also shell out to `shasum`.

We could also make a small babashka.crypto namespace (or some other name) with util functions.

Prior art: 
- https://github.com/funcool/buddy-core/blob/master/src/buddy/core/crypto.clj
- https://candid82.github.io/joker/joker.crypto.html

I'd rather not depend on / expose buddy directly since that would maybe bloat bb too much with stuff which isn't used that much.

Alternative: buddy pod. See #690

tjefferson08:
Thanks for keeping track of things, @borkdude 

Here's the java code that I wanted to be available from `java.crypto` when I cooked up the idea to shell out to openssl:

```java
import javax.crypto.Mac;
import javax.crypto.spec.SecretKeySpec;

public class HMAC {
  static public byte[] calcHmacSha256(byte[] secretKey, byte[] message) {
    byte[] hmacSha256 = null;
    try {
      Mac mac = Mac.getInstance("HmacSHA256");
      SecretKeySpec secretKeySpec = new SecretKeySpec(secretKey, "HmacSHA256");
      mac.init(secretKeySpec);
      hmacSha256 = mac.doFinal(message);
    } catch (Exception e) {
      throw new RuntimeException("Failed to calculate hmac-sha256", e);
    }
    return hmacSha256;
  }
}
```
(from [this site](https://sorenpoulsen.com/calculate-hmac-sha256-with-java))

Disclaimer: I don't know anything about crypto

Heliosmaster:
From a project where i did something similar, i used these classes:
```
 (:import (javax.crypto.spec SecretKeySpec)
           (javax.crypto Mac)
           (org.apache.commons.codec.digest DigestUtils))
```
and this code for hmac-256

```(defn hmac-sha256 [key string]
  "Returns the signature of a string with a given
    key, using a SHA-256 HMAC."
  (let [mac (Mac/getInstance "HMACSHA256")
[...]

PR #1066, opened 2021-11-15
Add javax.crypto classes necessary for calculating hmac shas.

grzm:
Used by was-api:

https://github.com/cognitect-labs/aws-api/blob/master/src/cognitect/aws/util.clj#L96-L100

Perhaps there's an alternative method to do this calculation with what's already in babashka. I haven't looked, and it's not an area I'm familiar with.


borkdude:
@grzm I think it's be good to add one or two tests that exercise these classes. Most notably we should make the examples mentioned in this issue work: https://github.com/babashka/babashka/issues/656.

PR #1077, opened 2021-11-26
Feature request to support shell execution of commands without an extension on MS-Windows

ikappaki:
Hi,

could you please consider patch to enable the tasks `shell` command execute MS-Windows commands without the need to supply an extension, à la the Command Prompt.

Currently, the `shell` cmd can't invoke any commands without supplying an extension, other than those which end in `.exe`.

Other than making the life of the user more difficult having to specify the extension, it affects cross platform compatibility with commands that are available both under the same name on *nix and MS-Windows, but the latter comes with a `.bat` or `.com` extension. Take Clerk's https://github.com/nextjournal/clerk/blob/main/bb.edn task file as an example file that I happen to stumble upon, that calls to `yarn` in`yarn-install`. Yarn is available under `yarn.cmd` on MS-Windows and any attempt to execute it with `shell` with fail with

```
Type:     java.io.IOException
Message:  Cannot run program "yarn": CreateProcess error=2, The system cannot find the file specified
Location: 21:19
```

It works fine if we call `yarn` from the Command Prompt (assuming it is in the PATH).

This means that tools like Clerk, whereby even their dev lifecycle tools are supposed to be cross platform (they are using babashka after all 😎 ), they are not actually no. I don't think it is up to them to correct the `bb.edn` file to check if they running are on MS-Windows so they can call `yarn.cmd` instead of `yarn`. 

As such, please find in the PR a proposal to support invocation of commands by `shell` without supplying an extension, following the same heuristics that Command Prompt is using as described in the `ext-add-windows` fn. 

The patch has all building blocks included, code, comments, tests but please feel free to scrutinize if you think is worth considering the patch. 

There are a couple of areas to be confirmed.

The first is about the `windows?` def, should the corresponding one in `babashka/process.clj` be made public instead? or should it move to a utility namespace?  There doesn't seem to be one.

The second is about the use of the `test-with-files` package that helps with testing a lot of the `ext-add-windows` functionality. Not sure if you mind having another test dependency in. Also that particular test assumes that the current working directory where the tests are invoked is the babashka repo root dir (where `lein test` has been invoked from) so that it can test relative and solo cmd executions. I don't think this is a problem, but just putting it out for being complete.

The `babashka.impl.tasks-test` tests pass on MS-Windows, and all tests pass on Windows.

I've tested the patch with the Clerk's bb tasks and it works.

Thanks!

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -77,6 +77,7 @@
                     :feature/selmer
                     :feature/logging
                     {:dependencies [[com.clojure-goes-fast/clj-async-profiler "0.5.0"]
+                                    [com.magnars/test-with-files "2021-02-17"]
                                     [com.opentable.components/otj-pg-embedded "0.13.3"]]}]
              :uberjar {:global-vars {*assert* false}
                        :jvm-opts ["-Dclojure.compiler.direct-linking=true"
--- a/src/babashka/impl/tasks.clj
+++ b/src/babashka/impl/tasks.clj
@@ -15,6 +15,76 @@
 (defn -chan? [x]
   (instance? ManyToManyChannel x))
 
+;; TODO copied over from src/babashka/process.clj
+(def windows?
+  (-> (System/getProperty "os.name")
+      (str/lower-case)
+      (str/includes? "windows")))
+
+(defn ext-add-windows
+  "Attempts to determine the actual filename extension that MS-Windows
+  Command Prompt would have used to locate and execute CMD.
+
+  The extension is
+
+  1. The extension of the CMD, if any.
+
+  2. or if CMD is an absolute path, then it is the first item in
+  ENV-PATHEXT list that when appended to CMD matches an existing file.
+
+  2. or if CMD is an relative path, then it is the first item in
+  ENV-PATHEXT list that when appended to CMD matches an existing file
+  relative to the current working directory.
+
+  3. or if CMD is just a solo filename, then it is the first item in
+  ENV-PATHEXT list that when appended to CMD matches an existing file
+  relative to the current working directory or relative to a path in
+  the ENV-PATH list (iterated from first to last).
+
+  ENV-PATH and ENV-PATHEXT default to the PATH and PATHEXT environment
+  variables respectively when not provided.
+
+  Returns CMD if it already came with an extension or an extension
+  could not be determined. Otherwise it returns CMD with the extension
+  appended."
+  ([cmd]
+   (ext-add-windows cmd (System/getenv "PATH") (System/getenv "PATHEXT")))
+  ([cmd env-path env-pathext]
+   (let [path (java.nio.file.Paths/get cmd (into-array String []))
+         ext? (-> path .getFileName .toString (str/includes? "."))]
+     (if ext?
+       cmd
+
+       (let [cmd-solo (.getFileName path)
+
+             paths (if (.isAbsolute path)
+                     ;; absolute
+                     [(-> path .getParent .toAbsolutePath .toString)]
+
+                     (if-let [parent (.getParent path)]
+                       ;; relative to current directory
+                       [(-> parent .toAbsolutePath .toString)]
+
+                       ;; solo filename, relative to CWD or PATH
+                       (-> (str ".;" env-path)
+                           (str/split #";")
+                           (->> (filter not-empty)))))
+
+             path-exts (when env-pathext
+                         (->> (str/split env-pathext #";")
+                              (filter not-empty)))]
+
+         (if-let [cmd-ext-path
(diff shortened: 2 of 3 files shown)


borkdude:
> I don't think it is up to them to correct the bb.edn file to check if they running are on MS-Windows so they can call yarn.cmd instead of yarn.

This is how I've been doing it on several occassions: https://github.com/borkdude/jayfu/blob/89b9a86fa9211d3f726b021fbe0549dc91c283ee/bb.edn#L25

One could ask why the JVM itself doesn't "make up" for this Windows behavior? Have you done any research to this?

I think `shell` is the wrong place to add this functionality. `shell` is just a wrapper around `babashka.process` and we should 

- probably add it either there
- or add a function to babashka.fs to find the Windows executable (there is already a function `which` which could do this)

A small detail: `fs` now has `with-tmp-dir` as well, so there is no need to bring in the test dependency.

I assume that @bobisageek may also have some ideas about this. 

I should probably make it more clear, but I don't think it's a good idea to start with a patch. Read here why:

https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#workflow

Rather, make an issue with a clear problem statement, alternatives considered, etc. Let's start from there and then we'll talk about the implementation.

ikappaki:
> > I don't think it is up to them to correct the bb.edn file to check if they running are on MS-Windows so they can call yarn.cmd instead of yarn.
> 
> This is how I've been doing it on several occassions: https://github.com/borkdude/jayfu/blob/89b9a86fa9211d3f726b021fbe0549dc91c283ee/bb.edn#L25

IMHO opinion is too much for a developer to consider every single idiosyncrasy of operating systems that they might have never used before. Things should work out of the box where possible.

> One could ask why the JVM itself doesn't "make up" for this Windows behavior? Have you done any research to this?

Yes, it is because they are simply calling down to the CreateProcess win32 API, which is the right thing to do when creating a process IMHO, they just let the underlying operating system deal with it

https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.base/windows/classes/java/lang/ProcessImpl.java#L580

https://docs.microsoft.com/en-gb/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw

taken from the above:

```
  If the file name does not contain an extension, .exe is appended. Therefore, if the file name extension is .com, this parameter must include the .com extension. If the file name ends in a period (.) with no extension, or if the file name contains a path, .exe is not appended. If the file name does not contain a directory path, the system searches for the executable file in the following sequence:

    The directory from which the application loaded.
    The current directory for the parent process.
    The 32-bit Windows system directory. Use the GetSystemDirectory function to get the path of this directory.
    The 16-bit Windows system directory. There is no function that obtains the path of this directory, but it is searched. The name of this directory is System.
    The Windows directory. Use the GetWindowsDirectory function to get the path of this directory.
    The directories that are listed in the PATH environment variable. Note that this function does not search the per-application path specified by the App Paths registry key. To include this per-application path in the search sequence, use the ShellExecute function.
```

> I think `shell` is the wrong place to add this functionality. `shell` is just a wrapper around `babashka.process` and we should
> 
>     * probably add it either there

IMHO the *shell* is the right place to put this into; the "shell" on windows is consider to be  the Command Prompt and it is the windows "shell"  (i.e. the Command Prompt) that does these translations for us when we use it, not the process API function. I understand that technically the `shell` command is wrapper around `babashka.process`, but I am viewing the `shell` something closer to that of a shell functionality (this is my main argument for having this into `shell`).

>     * or add a function to babashka.fs to find the Windows executable (there is already a function `which` which could do this)

ok, but I would like to exhaust all possible avenues pushing for the original suggestion first please :)

> A small detail: `fs` now has `with-tmp-dir` as well, so there is no need to bring in the test dependency.

sounds fantastic :)

> I assume that @bobisageek may also have some ideas about this.
> 
> I should probably make it more clear, but I don't think it's a good idea to start with a patch. Read here why:
> 
> https://github.com/clj-kondo/clj-kondo/blob/master/doc/dev.md#workflow
> 
> Rather, make an issue with a clear problem statement, alternatives considered, etc. Let's start from there and then we'll talk about the implementation.

A very good point indeed! I will raise a jira. I saw this as being a new feature rather than an issue, and that's why a jumped the gun directly to a PR.

Thanks for the quick reply!

borkdude:
A JIRA? I didn't know we had one :) A Github issue will do just fine :).

ikappaki:
> A JIRA? I didn't know we had one :) A Github issue will do just fine :).

Hehe indeed. raised #1078.

Issue #385, opened 2020-04-26
support for sqlite


borkdude:
There is a [sqlite](https://github.com/borkdude/babashka/tree/sqlite) branch based on information from this project: https://github.com/mageddo/graalvm-examples/tree/master/sqlite

I tried to port all the reflection config and jni config, but when I execute:

```
$ ./bb examples/sqlite_unused_vars.clj
```
it hangs, without printing any results.

Binary size grows with about ~20MB (double the size of hsqldb).

PR #1079, opened 2021-11-27
Add support for SQLite via a BABASHKA_FEATURE_SQLITE feature flag

nikvdp:

- [X] This PR correponds to an [issue with a clear problem statement](https://github.com/babashka/babashka/blob/master/doc/dev.md#start-with-an-issue-before-writing-code).

---

This PR corresponds to issue https://github.com/babashka/babashka/issues/385.

The hard part (getting [xerial/sqlite-jdbc](https://github.com/xerial/sqlite-jdbc) to run under GraalVM) has been taken care of by @ericdallo in [ericdallo/sqlite-jni-graal-fix](https://github.com/ericdallo/sqlite-jni-graal-fix). This PR just wires that work up as a babashka feature flag and updates [script/compile](https://github.com/babashka/babashka/compare/master...nikvdp:sqlite?expand=1#diff-00d7b856672fec3b9572542d459eb30bc2b8bfc69381f3635081811a5202196fR73) to include the prebuilt sqlite C libraries provided by xerial/sqlite-jdbc.

<details>
<summary>The following snippet can be used in <code>bb</code> as a quick test to verify that SQLite support is working as expected:</summary>

```clojure
(ns sqlite-test
  (:require [clojure.pprint :refer [pprint]]
            [next.jdbc :as ndbc]
            [next.jdbc.result-set :as rs]))

(def db {:dbtype "sqlite" :dbname  ":memory:"})
(defn nquery [db query]
  (let [query (if (string? query) [query] query)]
    (ndbc/execute! db query {:builder-fn rs/as-unqualified-lower-maps})))

(pprint (nquery db "select 'hello from sqlite at ' || time() as hello;"))
;; [{:hello "hello from sqlite at 06:14:51"}]
```
</details>

I have not tested this on Windows, but I followed the feature flag checklist laid out [here](https://github.com/babashka/babashka/commit/13f65f05aeff891678e88965d9fbd146bfa87f4e) and updated the windows .bat file accordingly, so hopefully it works there as well.

On macOS enabling this feature flag increases the `bb` binary size by about ~13MB and on Linux by about ~18MB.

A few questions:
1. The resulting `bb` binary can only actually use SQLite if both `export BABASHKA_FEATURE_JDBC=true` *and* `BABASHKA_FEATURE_SQLITE=true` were set before doing `script/uberjar` and `script/compile`. I added a `:feature/jdbc` dependency in the lein profile for `feature/sqlite`, but that alone doesn't resolve the issue. Would it make sense to add some logic in `script/uberjar` and/or `script/compile` to automatically set `BABASHKA_FEATURE_JDBC=true` when `BABASHKA_FEATURE_SQLITE=true`?
2. I'd like to add some documentation and examples of how to build and use this, where would the most useful place to do that be? Perhaps somewhere in [projects.md](https://github.com/babashka/babashka/blob/master/doc/projects.md)?
3. The second commit is not strictly SQLite related, it just exposes next.jdbc's result-set namespace to babashka. Would it be better to issue that as a separate PR?


borkdude:
@nikvdp Thanks! A few remarks:

- I think sqlite already includes the native image config so there is no need to include @ericdallo's older library for this: https://github.com/xerial/sqlite-jdbc/commits/master/src/main/resources/META-INF/native-image/org.xerial/sqlite-jdbc

- The resource config should in theory already be covered by https://github.com/xerial/sqlite-jdbc/blob/master/src/main/resources/META-INF/native-image/org.xerial/sqlite-jdbc/resource-config.json

- Please provide the adding of the namespace as a separate issue/PR (please make an issue first explaining why you want to add it)

- Let's keep the features separate and not add a dependency between the two, although commonly you will use them together. Just mention in the docs that you need to include both if you want to use next.jdbc + sqlite.

nikvdp:
Awesome, that's much simpler! I've bumped up the xerial/jdbc-sqlite version to pull in those changes and removed the older ericdallo fix and resource configuration. Also removed the next.jdbc result-set namespace stuff from this branch (will create a new issue and PR for that).

Tested on linux and mac again and all seems to be working as expected

ericdallo:
I'll add a note to the fix repro to mention that is not needed anymore :)

PR #1099, opened 2021-12-09
Build arm64 docker image in CI

### LINKED ISSUE #1097: Build & publish linux/arm64 Docker image
Once #1095 or similar lands, it would be nice to have the CI/CD pipeline use it to build and publish a linux/arm64 Docker image to Docker Hub in addition to the linux/amd64 already there.

Mostly opening this issue because I'm already working on this locally while waiting for #1095's CI infrastructure issues to settle down. :)

I have the [modified actions running in my fork](https://github.com/cap10morgan/babashka/actions) and I'll open a PR here when everything is working.

cap10morgan:

- [X] This PR correponds to an [issue with a clear problem statement](https://github.com/babashka/babashka/blob/master/doc/dev.md#start-with-an-issue-before-writing-code): Closes #1097 

- [ ] This PR contains a [test](https://github.com/babashka/babashka/blob/master/doc/dev.md#tests) to prevent against future regressions - I didn't touch the code itself, just CI infrastructure stuff.


lispyclouds:
Thanks for this @cap10morgan !

@borkdude did you see this error?
```
Apologies! Your build didn’t run because you haven’t selected a plan that includes macOS yet. Please change to a Performance Plan. If you have any questions please feel free to contact support.
```
should something be need to be done on Circle?

Also not sure why only the pulling of lambdaisland/regal fails?

lispyclouds:
Ah i guess its because its running on @cap10morgan 's fork?

borkdude:
Apparently. If @cap10morgan unsubscribes this repo from his personal CircleCI account then we can run it in ours.

cap10morgan:
@borkdude @lispyclouds I have "unfollowed" my fork in CircleCI. Let me know if I need to do anything else in there.

borkdude:
Now make another dummy commit.

lispyclouds:
Looks good! I guess the way to pull images after this would be `babashka/babashka:<version>-linux-arm64` or `babashka/babashka:<version>-linux-aarch64` right? Could this be improved by using buildx somehow? if someone downloads without the platform tag, docker selects automatically? I imagine setting up buildx isnt that straightforward on linux? Or explicit tags are preferred? my feeling is if we can leave out the platform from the tag, could be nicer to support more things maybe later? what's your opinion on it too @borkdude ?

cap10morgan:
> Looks good! I guess the way to pull images after this would be `babashka/babashka:<version>-linux-arm64` or `babashka/babashka:<version>-linux-aarch64` right? Could this be improved by using buildx somehow? if someone downloads without the platform tag, docker selects automatically? I imagine setting up buildx isnt that straightforward on linux? Or explicit tags are preferred? my feeling is if we can leave out the platform from the tag, could be nicer to support more things maybe later? what's your opinion on it too @borkdude ?

Not quite / the improvement you seek is already there.

Here are the nitty gritty details:

The platform tags are currently necessary because we're building each platform on a separate VM in CI. That means we have to use a docker manifest to combine them into one tag post-build. And we have to push them to the registry first before the manifest can find them and combine them.

Once the manifest is pushed, the tags you're already using will have both platforms and Docker will use the best one for your system by default. You can ignore (delete?) the tags with the platform names in them. They are just temporary build artifacts.

I would also like to skip cluttering the registry with those platform tags, but the way to do that isn't obvious in a CI environment. Your choices are either: 1. Use qemu emulation to execute a `docker buildx build --platform linux/amd64,linux/arm64 --push ...` command (very, very slow to the point of unusability where GraalVM native-images are concerned) or 2. Setup a Docker build farm w/ context nodes added to your buildx builder instance and farm out each platform to a native node. I just got this working locally and it's pretty nice once you get all the pieces together, but I'm not sure how feasible it would be to do in a CI environment like CircleCI or GitHub Actions w/o self-hosting it.

But maybe we can delete the platform tags at the end of the build? I can look into that if desired.

cap10morgan:
Actually, since we're doing all the heavy lifting (i.e. GraalVM native-image building) outside of Docker anyway, it might be feasible to combine the Docker image build & push into one buildx command and skip these interim platform tags. Since we're just wrapping the `bb` binary at the docker image stage, emulation might not be so bad.

I'll take a crack at it...

borkdude:
I'm fine with whatever choice you guys end up making as long as the current images aren't broken (no breaking changes for consumers).

cap10morgan:
> I'm fine with whatever choice you guys end up making as long as the current images aren't broken (no breaking changes for consumers).

Agreed, and they are not

lispyclouds:
Thanks for the details @cap10morgan ! Yeah i was gonna suggest if we can consume the binaries from the build and then make the docker images

cap10morgan:
TODO before merging:

1. Add non-PR & master branch guard back into `.circleci/script/docker`
2. Remove PR-specific branches from docker job branch filter in `.circleci/config.yml`

cap10morgan:
OK this is working now without creating the intermediate platform tags. It does this by building the thin wrapper images for both platforms at once (using buildx) and pushing those to Docker Hub.

I'll go through the checklist of pre-merge TODOs now and stop following my fork in CircleCI again.

.circleci/config.yml
```diff
@@ -28,13 +36,13 @@ jobs:
...
           command: |
             sudo script/install-clojure
-      - run:
-          name: Run JVM tests
-          command: |
-            export BABASHKA_FEATURE_JDBC=true
-            export BABASHKA_FEATURE_POSTGRESQL=true
-            script/test
-            script/run_lib_tests
+      # - run:
```
lispyclouds, line 39: we can uncomment this now right?

.circleci/script/docker
```diff
@@ -4,6 +4,7 @@ set -eo pipefail
 
 image_name="babashka/babashka"
 image_tag=$(cat resources/BABASHKA_VERSION)
+platform=${PLATFORM:-"linux/amd64"}
```
lispyclouds, line 7: since this is the only place we use the `PLATFORM` variable, better to inline define it here than an env var? less of a context switch to see where its defined i think.

lispyclouds:
The docker script is sufficiently nash-magical enough for me to warrant a re-write in bb sometime! great bash skills @cap10morgan ! I defintely had to search how some of what you wrote works 😛

.circleci/script/docker
cap10morgan: I'm relying on it being an env var here: https://github.com/babashka/babashka/blob/79e9bf2ba5f2f02e7c45e56a18ac210fe779872d/.circleci/config.yml#L462

...so that if you just run this script w/o overriding that var, it defaults to just `linux/amd64` like it does on `master`. But if you'd rather I just set it to `linux/arm64,linux/amd64` here, I'm happy to do that.

cap10morgan:
> The docker script is sufficiently bash-magical enough for me to warrant a re-write in bb sometime! great bash skills @cap10morgan ! I defintely had to search how some of what you wrote works 😛

👍🏻 👍🏻 to that bb rewrite idea! When it comes to bash, familiarity definitely breeds contempt. :)

.circleci/script/docker
lispyclouds: yeah makes sense, i missed the defaults that the script takes when the var is not changed. looks good to me.

.circleci/config.yml
```diff
@@ -35,10 +43,10 @@ jobs:
             export BABASHKA_FEATURE_POSTGRESQL=true
             script/test
             script/run_lib_tests
-      # - run:
-      #     name: Run as tools.deps dependency
-      #     command: |
-      #       .circleci/script/tools.deps
+      - run:
```
lispyclouds, line 46: This is needed now? @borkdude im not sure what is it used for?

borkdude: We can probably leave this out

lispyclouds: awesome! @cap10morgan would you delete/comment it?

borkdude: Isn't it already commented?

lispyclouds:
Looks good! Thanks for this!

PR #1106, opened 2021-12-14
Document babashka forks

cldwalker:
Hi @borkdude. Following up on documenting spec.alpha. I figured I'd document here since the book already links here. Also removed double nesting of Projects in the table of contents


borkdude:
Thank you!

PR #1112, opened 2021-12-21
Add libtest

cldwalker:
Hi @borkdude. I followed up on my comment on #1109 and added a script to make testing new libraries easy. There's an example in the documentation. With this script I added two new libraries and added tests for 3 existing libraries. I did have to make a couple minor fixes and they are noted with a "TEST-FIX" comment.

While trying to add tests for data.zip I encountered that `clojure.xml` which ships with clojure doesn't exist. It may be worth porting at some point. You can reproduce the failure with `script/add-libtest.clj '{org.clojure/data.zip {:mvn/version "1.0.0"}}' https://github.com/clojure/data.zip --test`

Cheers

- [-] I have updated the [CHANGELOG.md](https://github.com/babashka/babashka/blob/master/CHANGELOG.md) file with a description of the addressed issue.
  Changelog isn't usually updated for tested libs

### DIFF
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -92,6 +92,8 @@ If the library you want to add doesn't work with the script, you can manually do
 * Add an entry in `run_all_libtests.clj` to run the added test namespaces.
 * Run the tests `script/lib_tests/run_all_libtests NS1 NS2`
 
+Note: If you have to modify a test to have it work with bb, add an inline
+comment with prefix "TEST-FIX:" explaining what you did.
 
 ## Build
 
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -81,11 +81,16 @@ Test the native version:
 ## Tests for Libraries
 
 Babashka runs tests of libraries that are compatible with it through
-`script/run_lib_tests`. To add tests for a new library, do the following:
+`script/run_lib_tests`. To add tests for a new library that has a git repository
+and run them, use the script `add-libtest.clj` e.g. `script/add-libtest.clj
+'{listora/again {:mvn/version "1.0.0"}}' https://github.com/liwp/again --test`.
+
+If the library you want to add doesn't work with the script, you can manually do the following:
 
 * Add an entry for the library in `deps.edn` under the `:lib-tests` alias.
 * Create a directory for the library in `test-resources/lib_tests/` and copy its tests to there.
 * Add an entry in `run_all_libtests.clj` to run the added test namespaces.
+* Run the tests `script/lib_tests/run_all_libtests NS1 NS2`
 
 
 ## Build
--- a/test-resources/lib_tests/babashka/run_all_libtests.clj
+++ b/test-resources/lib_tests/babashka/run_all_libtests.clj
@@ -59,41 +59,13 @@
 (require '[cprop.source :refer [from-env]])
 (println (:cprop-env (from-env)))
 
-;;;; comb
-
-;; TODO: port to test-namespaces
-
-(require '[comb.template :as template])
-(prn (template/eval "<% (dotimes [x 3] %>foo<% ) %>"))
-(prn (template/eval "Hello <%= name %>" {:name "Alice"}))
-(def hello
-  (template/fn [name] "Hello <%= name %>"))
-(prn (hello "Alice"))
-
-;;;; arrangement
-
-;; TODO: port to test-namespaces
-
-(require '[arrangement.core :as order])
-(prn (sort order/rank ['a false 2 :b nil 3.14159
-                       "c" true \d [3 2] #{:one :two}
-                       [3 1 2] #{:three}]))
-
 ;;;; clj-yaml
 
 (test-namespaces 'clj-yaml.core-test)
 
-;;;; clojure-csv
-
-;; TODO: port to test-namespaces
-
-(require '[clojure-csv.core :as csv])
-;; TODO: convert to test
-(prn (csv/write-csv (csv/parse-csv "a,b,c\n1,2,3")))
-
 ;;;; clojure.data.zip
 
-;; TODO: port to test-namespaces
+;; TODO: port to test-namespaces. Blocked until clojure.xml is supported
 
 (require '[clojure.data.xml :as xml])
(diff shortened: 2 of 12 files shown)


doc/dev.md
```diff
@@ -81,12 +81,19 @@ Test the native version:
...
+'{listora/again {:mvn/version "1.0.0"}}' https://github.com/liwp/again --test`.
+
+If the library you want to add doesn't work with the script, you can manually do the following:
 
 * Add an entry for the library in `deps.edn` under the `:lib-tests` alias.
 * Create a directory for the library in `test-resources/lib_tests/` and copy its tests to there.
 * Add an entry in `run_all_libtests.clj` to run the added test namespaces.
+* Run the tests `script/lib_tests/run_all_libtests NS1 NS2`
 
+Note: If you have to modify a test to have it work with bb, add an inline
```
cldwalker, line 95: I didn't know if there was an existing convention so I made one. Useful to know modifications if we ever want to update tests

test-resources/lib_tests/bb-tested-libs.edn
```diff
@@ -0,0 +1,8 @@
+;; These libraries are tested against babashka and have been added by
```
cldwalker, line 1: If we're able to port most of the tested libraries to this config, this could be used to generate a nicer projects list

cldwalker:
If you find this new script and bb-tested-libs.edn to be handy, could I port existing library tests to use it?

borkdude:
> While trying to add tests for data.zip I encountered that clojure.xml which ships with clojure doesn't exist.

In discussions on Slack I've understood that `clojure.xml` is effectively deprecated. Given that I don't think it makes sense to port to babashka.

borkdude:
Can you explain more in words what this PR/script does? It clones a repo and then copies the tests?

cldwalker:
> In discussions on Slack I've understood that `clojure.xml` is effectively deprecated. Given that I don't think it makes sense to port to babashka.

Ah, didn't know. Makes sense

> Can you explain more in words what this PR/script does? It clones a repo and then copies the tests?

Sorry for the lack of clarity. This PR automates testing a library - it adds the deps.edn entry, copies tests, adds an entry to an edn file tracking tested libraries and optionally invokes the added tests. With this script I was able to add 5 tests for libraries. I chose to use an edn file for tested libraries as I think it allows for future functionality e.g. upgrading tests, generating a projects page, etc.

borkdude:
Sounds good!

> adds an entry to an edn file tracking tested libraries

Why is this necessary?

cldwalker:
> Why is this necessary?

It's not necessary but I think it allows for more functionality down the road. If all the tested libraries were in an edn file rather than codified in https://github.com/babashka/babashka/blob/master/test-resources/lib_tests/babashka/run_all_libtests.clj, we could easily have scripts to upgrade tests, analyze tested libraries for usage of a babashka feature, etc. I think it'd also allow for a nicer, more up to date projects.md page at some point. We could at least generate a csv of tested libraries so that github renders it in a [well formatted, searchable table](https://github.com/Caged/portland-census/blob/6c71d5317fb0fb36c78068b8fcb8e59d69f86bad/source/data/2010-census.csv)

borkdude:
But they are/should also be in `deps.edn` under the libtests alias?

cldwalker:
Maven info is but additional data like version that has been tested and git url aren't

test-resources/lib_tests/babashka/run_all_libtests.clj
```diff
@@ -61,41 +59,13 @@
...
-;; TODO: port to test-namespaces
-
-(require '[clojure-csv.core :as csv])
-;; TODO: convert to test
-(prn (csv/write-csv (csv/parse-csv "a,b,c\n1,2,3")))
-
 ;;;; clojure.data.zip
 
-;; TODO: port to test-namespaces
+;; TODO: port to test-namespaces. Blocked until clojure.xml is supported
```
borkdude, line 68: As mentioned, clojure.xml isn't going to be supported because it's effectively deprecated.

doc/dev.md
borkdude: Perhaps choose a more specific one like `BB-TEST-PATCH`: ...

test-resources/lib_tests/bb-tested-libs.edn
borkdude: Good idea, but perhaps we can also use this config to add it as `-Sdeps` in addition to the test alias, to not duplicate stuff in both files? We can use a fully qualified keyword for test-namespaces to not cause conflicts with any other tools.

test-resources/lib_tests/babashka/run_all_libtests.clj
cldwalker: Yep. Will revert

doc/dev.md
cldwalker: 👍

test-resources/lib_tests/bb-tested-libs.edn
cldwalker: Happy to make changes but want to make sure I understand

> Good idea, but perhaps we can also use this config to add it as -Sdeps in addition to the test alias, to not duplicate stuff in both files?

So add a :deps entry to this file and only place these dependencies here and not in the root deps.edn, right?

> We can use a fully qualified keyword for test-namespaces to not cause conflicts with any other tools
Does something like this work?
```clj
{listora/again
 {:git-sha "b1a6793f0deaa3cc016eee7626097ba8bf36ba10", :git-url "https://github.com/liwp/again", :bb-test-lib/test-namespaces (again.core-test)}
 ...
```

borkdude: I would use `:git/sha` and `:git/url` as otherwise these don't work with deps.edn.
And then an extra key `:org.babashka/test-namespaces [...]`

borkdude:
@cldwalker I went ahead and merged this, we can address other things as follow up issues. Maybe it's better to keep the deps.edn (tested libs) and tests (from other config) separate like you have now, I'm ok with that too.

Issue #1110, opened 2021-12-16
Add --bb-edn or --config option to specify other bb.edn file 


borkdude:
The other bb.edn (or bb.ci.edn, etc) file can live in another directory.
Relative paths in `:paths` should be resolved against the explicitly provided `--config`'s location.

In the future we can also add `--context` for setting the directory to which we should resolve relative dirs againt, but we can default to the explicit `--config` option when given, or the current working directory otherwise (like it is now).

See discussion in https://github.com/babashka/babashka/discussions/869

/cc @wilkerlucio @cldwalker @thiagokokada @lispyclouds

cldwalker:
I'd be in favor of suggested behavior above. The two use cases I have are:
* Given an explicit config file, run a babashka executable with relative paths resolved to explicit file e.g. [bb-github-repo](https://github.com/cldwalker/bb-clis/blob/master/bin/bb-github-repo) as an executable
* Given an explicit config file, run a babashka task with relative paths resolved to explicit file e.g. `bb --config CONFIG run FOO`

borkdude:
Also see #1103 which is somewhat related to invoking tasks from another tasks file, but with merge behavior.

cldwalker:
That does look helpful and related but it's different enough that I don't think it address my uses cases.  Since #1103 is task specific it doesn't address the executable use case. For the task use case, #1103 requires a local bb.edn in order to merge in another bb.edn. I'm looking to invoke another bb.edn from the commandline

borkdude:
@cldwalker Agreed. I hope to look into this issue soon (but feel free to experiment).

borkdude:
Should work now in branch `other-bb-edn`. Still have to write some tests and docs, but with the manual tests I did it worked.

The argument is `--config other-bb.edn`.  

macOS: https://24985-201467090-gh.circle-artifacts.com/0/release/babashka-0.7.1-SNAPSHOT-macos-amd64.tar.gz
linux: https://24986-201467090-gh.circle-artifacts.com/0/release/babashka-0.7.1-SNAPSHOT-linux-amd64-static.tar.gz
Windows:

borkdude:
Merged to master.

borkdude:
Btw I also implemented `--deps-root` for explicitly setting the dependency root for relative paths.

borkdude:
You can now install a dev release: https://github.com/babashka/babashka-dev-builds
This should hopefully help getting this tested!

cldwalker:
[...]

PR #1117, opened 2021-12-25
[#1110] Load tasks and deps from other bb.edn file

borkdude:


borkdude:
@bobisageek @cldwalker Somehow the Windows tests are failing, but I don't know why. Perhaps I introduced a related change but I'm not aware of any. Any ideas?

bobisageek:
Proximal cause: 
The failures appear to just be related to line endings. The Appveyor build of #1114 (https://ci.appveyor.com/project/borkdude/babashka/builds/41981367) actually shows the same two failures in the lib_test status.

So, a quick fix for making the tests pass is to generify `"\n"` to `(System/lineSeparator)` (or otherwise handle variable line endings) at https://github.com/babashka/babashka/blob/cd31b11a1bb1acd069b9b8da7da5025cdfcdbdbd/test-resources/lib_tests/clojure/data/json_test.clj#L416-L418 and https://github.com/babashka/babashka/blob/cd31b11a1bb1acd069b9b8da7da5025cdfcdbdbd/test-resources/lib_tests/table/core_test.clj#L9-L19

Deeper stuff:
Why did #1114 pass Appveyor and this one fail, given that they both had lib_test failures reported. This is 95% conjecture, but I would guess that before the changes made on the `other-bb-edn` branch, the exit code of the lib test batch file always ended up being 0 because the last thing was unsetting the BABASHKA_EDN variable, and Windows batch doesn't have (or at least, the batch file doesn't have setup) something like the *nix 'pipefail' construct.

Summary:
- A couple of places in the lib tests need to account for variable line endings - I can put in a PR for this (although it's only two little changes) - would you like that PR to the `other-bb-edn` branch so that CI can be re-run on this PR, or would you prefer a different tactic?
- The changes made on this branch should improve the capability of failing lib tests to fail the Appveyor build.

borkdude:
@bobisageek Thanks for looking into this! A PR to the branch seems good!

PR #1118, opened 2021-12-25
handle system-specific newlines in a couple of tests

bobisageek:
This just tweaks the line endings in the test comparisons for a couple of tests, in order to eliminate the Appveyor failures.


bobisageek:
Well, the line endings get straightened out and Appveyor passes... the failure on the static build seems like it might be related to an unexpected response from the `postman-echo` site. I think that could safely be ignored, or maybe re-run to ensure that it was just an intermittent external failure.

borkdude:
Thank you!

cldwalker:
@bobisageek Thanks for fixing errors I introduced! When doing testlib updates, I'll be on the lookout for future weirdness with Appveyor passing but still having failures

PR #1139, opened 2022-01-12
Add bill-of-materials to Docker images to aid vulnerability scanning

kipz:
Second attempt at https://github.com/babashka/babashka/pull/1135

This tries to add the matabom jar to all 3 docker image builds, but is hard to test without attempting a release on github/circle-ci

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,5 +73,6 @@ RUN ./script/compile
 FROM ubuntu:latest
 RUN apt-get update && apt-get install -y curl \
         && mkdir -p /usr/local/bin
+COPY --from=BASE /opt/target/metabom.jar /opt/babashka-metabom.jar
 COPY --from=BASE /opt/bb /usr/local/bin/bb
 CMD ["bb"]
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -24,6 +24,7 @@ RUN apk --no-cache add curl ca-certificates tar && \
     apk add --allow-untrusted /tmp/glibc-2.28-r0.apk
 RUN echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf
 
+COPY metabom.jar /opt/babashka-metabom.jar
 COPY --from=tester /bin/bb /bin/bb
 
 CMD ["bb"]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,7 +55,6 @@ jobs:
             VERSION=$(cat resources/BABASHKA_VERSION)
             jar=target/babashka-$VERSION-standalone.jar
             cp $jar /tmp/release
-
             java -jar $jar script/reflection.clj
             reflection="babashka-$VERSION-reflection.json"
             java -jar "$jar" --config .build/bb.edn --deps-root . release-artifact "$jar"
@@ -479,7 +478,7 @@ workflows:
       - docker:
           filters:
             branches:
-              only: 
+              only:
                 - master
           requires:
             - linux
--- a/.circleci/script/docker
+++ b/.circleci/script/docker
@@ -33,6 +33,8 @@ if [ -z "$CIRCLE_PULL_REQUEST" ] && [ "$CIRCLE_BRANCH" = "master" ]; then
         if [[ $tarball_platform == "linux-arm64" ]]; then tarball_platform="linux-aarch64"; fi
         mkdir -p $p
         tar zxvf "/tmp/release/babashka-${image_tag}-${tarball_platform}.tar.gz" -C $p
+        # this overwrites, but this is to work around having built the uberjar/metabom multiple times
+        cp "/tmp/release/${tarball_platform}-metabom.jar" ./metabom.jar
     done
     docker buildx build -t "$image_name:$image_tag" --platform "$platform" "${label_args[@]}" --push -f Dockerfile.ci .
     if [[ $snapshot == "false" ]]; then
--- a/.circleci/script/release
+++ b/.circleci/script/release
@@ -6,9 +6,7 @@ cp bb /tmp/release
 
 VERSION=$(cat resources/BABASHKA_VERSION)
 
-cd /tmp/release
-mkdir -p /tmp/bb_size
-./bb '(spit "/tmp/bb_size/size" (.length (io/file "bb")))'
+
 
 ## release binary as tar.gz archive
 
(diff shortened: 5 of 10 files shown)


Dockerfile
```diff
@@ -73,5 +73,6 @@ RUN ./script/compile
 FROM ubuntu:latest
 RUN apt-get update && apt-get install -y curl \
         && mkdir -p /usr/local/bin
+COPY --from=BASE /opt/target/metabom.jar /opt/babashka-metabom.jar
```
borkdude, line 76: Where is `/opt/target` created?

kipz: This is one of the few bits I could test a bit locally: `/opt` is the `WORKDIR` for the Dockerfile, and `target/metabom.jar` is created by `lein metajar`

borkdude: Oh right, this is Dockerfile and not Dockerfile.ci, yeah, I can see how this was confusing in the previous PR :)

Dockerfile.alpine
```diff
@@ -24,6 +24,7 @@ RUN apk --no-cache add curl ca-certificates tar && \
     apk add --allow-untrusted /tmp/glibc-2.28-r0.apk
 RUN echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf
 
+COPY /tmp/release/metabom.jar /opt/babashka-metabom.jar
```
borkdude, line 27: @lispyclouds How is the release artifact mounted in the Docker container again? It would surprise me if /tmp/release should be used, given `COPY bb /bin/bb` above.

kipz: I _think_ it's available in the build-context for the docker images e.g.
https://github.com/kipz/babashka/blob/3116158fdacdcf84f9c583269d71c687865dd615/.circleci/script/docker#L37
It looks like the circle-ci yamls make this available to these steps, and it's the tar.gz of babashka that gets used. It's just a bit confusing because the bb we just built gets used during the build, archived, and then removed.

borkdude: Ah ok. I'll let @lispyclouds have a look and merge this.

lispyclouds: @kipz its a bit complex but the working directory when the docker build/x happens is the place where the source is cloned, proabably `~/project`, where circleci clones things by default. So the COPY from /tmp/release wont work as its outside the build context.

One way to do this could be to copy the metabom.jar from /tmp/release to the pwd like we extract the [tar](https://github.com/babashka/babashka/blob/master/.circleci/script/docker#L35) into the linux/amd64 directorty. The copy could happen in the docker script and then change the COPY in the dockerfiles to use the relative path to pick the jar up.

borkdude: I'd say, let's just test in this branch by enabling the Docker build and merge when things have been found to work.

lispyclouds: If you see the docker [job](https://github.com/babashka/babashka/blob/master/.circleci/config.yml#L447) in the circle yaml, we dont really change the pwd after the clone and hence its not where you were expecting i guess.

lispyclouds:
@kipz Also a general observation like @borkdude asked, would a deps tool be available for metabom too? We may have a plan to move to tools.deps completely in the future.

kipz:
> @kipz Also a general observation like @borkdude asked, would a deps tool be available for metabom too? We may have a plan to move to tools.deps completely in the future.

It's such a super simple plugin, so I'm happy to make it work for tools.deps if needed for sure.

borkdude:
@kipz I think you don't need to make a plugin, but just a library which you can use from a `build.clj` file.

Dockerfile.alpine
lispyclouds: @kipz you can remove https://github.com/babashka/babashka/blob/master/.circleci/config.yml#L480-L483 in your fork to test out the docker builds.

kipz: @lispyclouds yeah, I saw that, but don't all the previous tasks put stuff in /tmp/releases - otherwise how does the tar.gz get in there? - like you say though, this is pretty complicated, and trying to understand it fully without being able to run/debug it makes this pretty tricky.

lispyclouds: yes all of them are in `/tmp/releases` but if you see the docker build commands, the context is `.` and they are running in the place where the project is cloned, hence we can say `-f Dockerfile.ci`. So the path `/tmp/releases` would be outside the build context and the COPY wont work inside the dockerfiles.

lispyclouds: The tar from /tmp/releases is actually extracted within the build context [here](https://github.com/babashka/babashka/blob/master/.circleci/script/docker#L35) and its no more read from /tmp/releases

lispyclouds: All these complexity arose from being able to do multi arch build, hopefully we can do a simpler refactor sometime. Apologies for the complexity! 😞

kipz: That's just how things like this evolve! Hopefully we can rewrite the whole thing using `bb` at some point :)

borkdude:
@lispyclouds @kipz Is this reviewed and/or mergeable?

lispyclouds:
Not yet complete, needs the changes I have mentioned about moving the jar from `/tmp/releases` to the build context are pending.

kipz:
I should be able to sort that in the next day or so @borkdude

kipz:
@lispyclouds this isn't pretty, but I've tested this as best I can and the docker builds seem to work. 

Two things I had to work around, which unfortunately has made things even more complicated:

1. The test step clears out `target`, so it wasn't available in the release step. It looks like `lein clean` was added specifically for some reason, so I didn't want to remove that.
2. Circle CI doesn't like the same file (the metabom jar) being added to the workspace multiple times, so I've used the release step to call it something different for each platform and then the docker step to choose one

I couldn't really think of a simpler way to get this working, especially the second problem, without a more substantial refactoring of the Circle CI build. Most of what happens in all the jobs is the same - or at least produce the same uberjar/metabom jar, and run many of the same tests, so we could factor that out into its own job and reuse it.

lispyclouds:
Looks much better to me now, thanks for the efforts @kipz ! We would try to revisit and simplify things sometime soon and hopefully in clojure!

lispyclouds (approved):

lispyclouds:
@kipz for future work, simply push more commits and not force push them, makes it easier for us to follow along, review and make some suggestions if needed! we would squash and merge anyways later 😄

kipz:
@lispyclouds thanks for being patient - happy to share my journey next time, though at times it was like a random walk to discover what was going on :)

lispyclouds:
Well given the complexity we have in these scripts the random walks are enlightening for us too!

PR #1141, opened 2022-01-13
Build alpine Docker images in CI when linux/amd64 is in platform list

### LINKED ISSUE #1140: alpine Docker images are no longer being published
The last alpine Docker images to be published were for 0.7.0. I broke this in my arm64 Docker image changes in #1099. Just creating this issue to reference from the PR I'm about to open.

cap10morgan:

- [X] This PR corresponds to an [issue with a clear problem statement](https://github.com/babashka/babashka/blob/master/doc/dev.md#start-with-an-issue-before-writing-code). Fixes #1140 

- [ ] This PR contains a [test](https://github.com/babashka/babashka/blob/master/doc/dev.md#tests) to prevent against future regressions - N/A

### DIFF
--- a/.circleci/script/docker
+++ b/.circleci/script/docker
@@ -45,7 +45,7 @@ if [ -z "$CIRCLE_PULL_REQUEST" ] && [ "$CIRCLE_BRANCH" = "master" ]; then
 
     # alpine doesn't provide upstream arm64 images yet
     if [[ $platform == *"linux/amd64"* ]]; then
-        tar zxvf  "/tmp/release/babashka-${image_tag}-${platform}-static.tar.gz"
+        tar zxvf  "/tmp/release/babashka-${image_tag}-linux-amd64-static.tar.gz"
         docker buildx build -t "$image_name:alpine" --platform=linux/amd64 "${label_args[@]}" -f Dockerfile.alpine .
         rm -f bb
         docker tag "$image_name:alpine" "$image_name:$image_tag-alpine"
--- a/.circleci/script/docker
+++ b/.circleci/script/docker
@@ -44,9 +44,9 @@ if [ -z "$CIRCLE_PULL_REQUEST" ] && [ "$CIRCLE_BRANCH" = "master" ]; then
     done
 
     # alpine doesn't provide upstream arm64 images yet
-    if [[ $platform == "linux-amd64" ]]; then
+    if [[ $platform =~ linux/amd64 ]]; then
         tar zxvf  "/tmp/release/babashka-${image_tag}-${platform}-static.tar.gz"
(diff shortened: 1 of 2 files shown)


.circleci/script/docker
```diff
@@ -44,9 +44,9 @@ if [ -z "$CIRCLE_PULL_REQUEST" ] && [ "$CIRCLE_BRANCH" = "master" ]; then
     done
 
     # alpine doesn't provide upstream arm64 images yet
-    if [[ $platform == "linux-amd64" ]]; then
+    if [[ $platform =~ linux/amd64 ]]; then
```
borkdude, line 47: What does this syntax mean?

cap10morgan: actually IIRC there's a simpler substring thingy I can use instead

borkdude: Where does the `-` -> `/` change come from?

cap10morgan: changed to substring match

cap10morgan: > Where does the `-` -> `/` change come from?

That should have already been in there, but it comes from the fact that docker platform strings look like `os/arch` while GraalVM and babashka use `os-arch`. There was a time where I was using the hyphenated form in this env var, but that was no longer the case by the time the original PR was merged and I just missed converting this one b/c it was never being run anyway.

lispyclouds: more reasons for this to be in clojure! quite some arcane bash magic!

lispyclouds:
@cap10morgan Im thinking do we need this if check at all? this script is only called once in the ci build process and the value of platform never really changes.

.circleci/script/docker
```diff
@@ -44,9 +44,9 @@ if [ -z "$CIRCLE_PULL_REQUEST" ] && [ "$CIRCLE_BRANCH" = "master" ]; then
     done
 
     # alpine doesn't provide upstream arm64 images yet
-    if [[ $platform == "linux-amd64" ]]; then
+    if [[ $platform == *"linux/amd64"* ]]; then
         tar zxvf  "/tmp/release/babashka-${image_tag}-${platform}-static.tar.gz"
```
lispyclouds, line 48: we should hardcode the `platform` to `linux-amd64` here right?

cap10morgan:
> @cap10morgan Im thinking do we need this if check at all? this script is only called once in the ci build process and the value of platform never really changes.

I'm happy to take it out if you think it's not needed. All this stuff was platform-parameterized when I found it, so I left it like that from a "first, do no harm" perspective. I'm fine w/ it either way.

lispyclouds:
Yeah lets remove it. We have no other use for the docker script apart from the CI, people building locally anyways use a different Dockerfile and proabably a manual process.

lispyclouds (approved):

PR #1142, opened 2022-01-13
Fix: alpine docker builds in CI (follow-up)

cap10morgan:
Follow up to #1141 that makes alpine tagging and pushing consistent with other variants

### DIFF
--- a/.circleci/script/docker
+++ b/.circleci/script/docker
@@ -49,9 +49,8 @@ if [ -z "$CIRCLE_PULL_REQUEST" ] && [ "$CIRCLE_BRANCH" = "master" ]; then
     docker buildx build -t "$image_name:$image_tag-alpine" --platform=linux/amd64 "${label_args[@]}" --push -f Dockerfile.alpine .
     rm -f bb
     if [[ $snapshot == "false" ]]; then
-        docker tag "$image_name:$image_tag-alpine" "$image_name:alpine"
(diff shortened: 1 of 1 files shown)


.circleci/script/docker
```diff
@@ -45,14 +45,12 @@ if [ -z "$CIRCLE_PULL_REQUEST" ] && [ "$CIRCLE_BRANCH" = "master" ]; then
 
     # build alpine image for linux-amd64 only (no upstream arm64 support yet)
     tar zxvf  "/tmp/release/babashka-${image_tag}-linux-amd64-static.tar.gz"
-    docker buildx build -t "$image_name:alpine" --platform=linux/amd64 "${label_args[@]}" -f Dockerfile.alpine .
+    echo "Building & pushing Docker image $image_name:$image_tag-alpine"
+    docker buildx build -t "$image_name:$image_tag-alpine" --platform=linux/amd64 "${label_args[@]}" --push -f Dockerfile.alpine .
```
lispyclouds, line 49: should we put this in the else branch of the if? this would be building twice if not a snapshot?

cap10morgan: It doesn't really build twice b/c it still caches the intermediate layers. We still want this tag pushed even when it's not snapshot, right?

lispyclouds: ah yes makes sense.

Issue #1143, opened 2022-01-13
`--` included in arguments to `-main` function when invoking `bb --main ...`


pjt:
**version**

\>= 0.6.5

**platform**

macOS, Linux, Docker

**problem**

In an invocation like `bb --main some.ns -- arg1 arg2`, `bb` passes `--` as the first argument to the `some.ns/-main`. This is a behavior change from versions before 0.6.5, where the `-main` function gets passed only the arguments following `--`.

**repro**

Versions at 0.6.5 and after below illustrate the bug.

```
$ cat f.clj
(ns f)

(defn -main [& args] (prn args))

$ cat test.sh
#!/bin/sh
VERSION=${1:-0.7.1}
echo $VERSION

docker run -v `pwd`:`pwd` -w `pwd` --rm -it babashka/babashka:${VERSION} bb -cp . --main f -- hi

$ for version in 0.3.0 0.4.0 0.5.0 0.6.0 0.6.1 0.6.2 0.6.3 0.6.4 0.6.5 0.6.6 0.6.7 0.6.8; do sh test.sh $version; done
0.3.0
("hi")
0.4.0
("hi")
0.5.0
("hi")
0.6.0
("hi")
0.6.1
("hi")
0.6.2
("hi")
0.6.3
("hi")
0.6.4
("hi")
0.6.5
("--" "hi")
0.6.6
("--" "hi")
0.6.7
("--" "hi")
0.6.8
("--" "hi")
```
**expected behavior**

Versions before 0.6.5 and after below illustrate the expected behavior.

PR #1147, opened 2022-01-20
Allow optional (ignored) -- when using using --main

grzm:
Permits previously-accepted behavior broken in 19415f63632d2fe8f668fa82ea91c80d06666292

Fixes https://github.com/babashka/babashka/issues/1143

- [X] This PR contains a [test](https://github.com/babashka/babashka/blob/master/doc/dev.md#tests) to prevent against future regressions

Nope. Not yet.

The added test for `--` as the next arg is simple and works but likely there's a more elegant/idiomatic solution.

### DIFF
--- a/src/babashka/main.clj
+++ b/src/babashka/main.clj
@@ -601,7 +601,9 @@ Use bb run --help to show this help output.
           ("--main", "-m",)
           (let [options (next options)]
             (assoc opts-map :main (first options)
-                   :command-line-args (rest options)))
+                   :command-line-args (if (= "--" (second options))
+                                        (nthrest options 2)
+                                        (rest options))))
           ("--run")
           (parse-run-opts opts-map (next options))
           ("--tasks")
--- a/test/babashka/main_test.clj
+++ b/test/babashka/main_test.clj
@@ -59,7 +59,8 @@
     (is (:babashka/version v))
     (is (:feature/xml v)))
   (is (= {:force? true} (parse-opts ["--force"])))
-  (is (= {:main "foo", :command-line-args '("-h")} (parse-opts ["-m" "foo" "-h"]))))
+  (is (= {:main "foo", :command-line-args '("-h")} (parse-opts ["-m" "foo" "-h"])))
+  (is (= {:main "foo", :command-line-args '("-h")} (parse-opts ["-m" "foo" "--" "-h"]))))
 
 (deftest version-test
   (is (= [1 0 0] (main/parse-version "1.0.0-SNAPSHOT")))


borkdude:
@grzm I'm not sure why we need this fix anymore. Although previously the `--` was eaten from the main args, should we? Is there a reason we would ever need `--` with main? Previously we did, but I don't think it's true anymore. Are there any downsizes to consuming `--` anyway? clojure.main also doesn't do it. I would like to hear your thoughts on this.

grzm:
> @grzm I'm not sure why we need this fix anymore. Although previously the `--` was eaten from the main args, should we? Is there a reason we would ever need `--` with main? Previously we did, but I don't think it's true anymore. Are there any downsizes to consuming `--` anyway?

The motivating factor and biggest downside of no longer consuming `--` after main is that it breaks backwards compatibility: we had failures in babashka scripts when we upgraded past 0.6.4 as we have relied on the behavior of `--main` consuming `--`: indeed, it was required.

I agree that the inconsistency with the behavior of `clojure --main` is unfortunate, and perhaps this is enough of a motivation to make this breaking change. As for other scripts supporting `--` for their own CLI, I don't know what typical behavior is in cases like this but I would be very surprised if there aren't already best practices or generally accepted behaviors of how to handle this.

Sleeping on it, I do think that the current patch might not be exactly right: `--` typically means "stop consuming args and pass the rest of arguments to the script". I think it's worth examine the current behavior of `babashka.main/parse-args` to see how this is handled generally— if `--` is to be supported at all. For backwards compatibility, I think consuming an initial `--` following `--main` is warranted.

borkdude:
If it's important to you, I will merge this and you can sleep another night for a more general solution.

grzm:
> If it's important to you, I will merge this and you can sleep another night for a more general solution.

I appreciate the kindness, but we're fine running on older versions of babashka for now. There's no need to merge at this point. I'm more interested in arriving at a solution you're happy with for the future of babashka.

borkdude:
@grzm What about upgrading and removing `--` after the `-m` invocation?

grzm:
> @grzm What about upgrading and removing `--` after the `-m` invocation?

If that's the direction you want to go, that's what we'll end up doing. We use babashka (and --main) extensively, so it won't be a trivial project, but we'll make the changes to keep up with babashka and features in newer releases that we'll want to leverage.

borkdude:
I think your argument of backwards-incompatibility is a strong one and if you have to make significant changes to your code, then I would say, let's fix it. I'll be happy to merge your PR as is and with an optional follow up.

grzm:
Cheers!

PR #1176, opened 2022-02-12
Ensure leading zeros are not trimmed for algorithm "SHA-256", so that…

rlhk:
… hash string size is always 64.


borkdude:
@rlhk What was the reason for this PR?

rlhk:
> @rlhk What was the reason for this PR?

As commit comment says, the newly added test case will fail by having its leading zero trimmed if we format with "%032x" for SHA-256. The resulting string is 63 characters long instead of the correct length of 64. So "%064x" was added to address the issue. Other algorithms will default to use "%032x". The map def in let block of the signature fn could add other sizes if other algorithms need different output string length.

Found this issue when I try to use the same signature fn for my project.

Issue #863, opened 2021-05-22
pods in bb.edn


borkdude:
We can have a pod config in bb.edn:

```
{:pods {org.babashka/mysql {:pod/version "0.0.6"}}}
```

When bb starts, we will check if all pods are already available. 

To load the pod, you no longer have to provide the version:

```
(load-pod 'org.babashka.mysql)
```

As a second step, we could list the namespaces when the pod should be loaded in bb.edn?

In the case of creating an uberjar, we'll have to remember the pod versions in order to be able to load them. Maybe insert the pod versions into the manifest?

FieryCod:
@borkdude Maybe it would be easier to just include bb.edn in uberjar instead?

PR #1205, opened 2022-03-10
Load pods from bb.edn

cap10morgan:

- [X] This PR corresponds to an [issue with a clear problem statement](https://github.com/babashka/babashka/blob/master/doc/dev.md#start-with-an-issue-before-writing-code). #863 


cap10morgan:
I'm sure this is far too naive and breaks in all kinds of situations, but wanted to get a draft up to start discussing what else it needs.

With this change I can require a pod in my `bb.edn` like this:

```clojure
{:paths ["src"]
 :pods  {com.fluree/crypto  {:registry/version "0.1.1"}}}
```

and then in my main ns just do this:

```clojure
(ns bb.declpods.core
  (:require [pod.fluree.crypto :as crypto])
  (:gen-class))

(defn greet
  "Callable entry point to the application."
  [data]
  (let [keys (crypto/generate-key-pair)]
    (println (str "Hello, " (or (:name data) "World") "!"))
    (println "Also I made you this:" (pr-str keys))))

(defn -main
  "I don't do a whole lot ... yet."
  [& args]
  (greet {:name (first args)}))
```

And when I run `bb -m bb.declpods.core` I get this:

```
Hello, World!
Also I made you this: {:private "8cd30c54bd7801ce71bb9405969bda654e6359ee80d334546be2870b953d720d", :public "0205bfe72aab2338095df72f956e2402cf02240d48463086bd538c04337781d51e"}
```

cap10morgan:
Known TODOs:

1. As discussed in #863, needs to work in uberjars. I lean towards pulling the `bb.edn` file into the uberjar personally.
2. Add more here...

borkdude:
```
:pods  {com.fluree/crypto {:registry/version "0.1.1"}}
```

I think we could just do: `:version` since that's what we have in `load-pod` already.

At 2: I think we should lazily load them, but to lazily load them we need to know the namespaces that they expose in advance. We could make that part of the pod manifest in the registry, but then we would need to update all the manifests. We could also call the pod once on installation and query all the namespaces (which is part of the describe message pods send). We could persist that information in the ~/.babashka/pods/repository` directory alongside the manifest and binaries and then use that information in `:load-fn`.

At 5. Instead I think we should make the pod registry more configurable so people can add their own registries. E.g. the fluree repo could host its own manifests and link to the binaries in releases.

cap10morgan:
> ```
> :pods  {com.fluree/crypto {:registry/version "0.1.1"}}
> ```
> 
> 
>      
>   
> 
> I think we could just do: `:version` since that's what we have in `load-pod` already.

You don't think there's another kind of version we might want to support later? Or maybe named registries where the name is the namespace in front of version here? Happy to do whichever you prefer here, just want to make sure.

> 
> 
> At 2: I think we should lazily load them, but to lazily load them we need to know the namespaces that they expose in advance. We could make that part of the pod manifest in the registry, but then we would need to update all the manifests. We could also call the pod once on installation and query all the namespaces (which is part of the describe message pods send). We could persist that information in the ~/.babashka/pods/repository`directory alongside the manifest and binaries and then use that information in`:load-fn`.

I like that idea of calling it once to load & cache the describe info. I'll look into that.

> 
> 
> At 5. Instead I think we should make the pod registry more configurable so people can add their own registries. E.g. the fluree repo could host its own manifests and link to the binaries in releases.

Yeah, that seems right. Upon thinking about it further, I think this stuff should go into a subsequent PR. It's only tangentially related to this one.

borkdude:
> You don't think there's another kind of version we might want to support later?

If we need to, we can support other things later, but we don't support it in load-pod either, which should probably also support those.

The order of the custom registries stuff probably will determine where the pod will be resolved from first.

src/babashka/main.clj
```diff
@@ -737,7 +737,9 @@ Use bb run --help to show this help output.
             _ (when jar
                 (cp/add-classpath jar))
             load-fn (fn [{:keys [:namespace :reload]}]
-                      (or (when-let [{:keys [:loader]}
+                      (or (when-let [pod (-> @common/ctx :env deref :pod-namespaces (get namespace))]
```
borkdude, line 740: Should we prioritize pod namespaces over source namespaces? Maybe yes... no sure. 
Perhaps we could have a faster check, when there aren't any pod namespaces, we can skip this check. As commented in the pods PR, I think making load-metadata just have return some data, we could store that data in a separate atom or volatile in bb.main and then check the data in that thing.

src/babashka/main.clj
```diff
@@ -687,6 +687,8 @@ Use bb run --help to show this help output.
 
 (def env (atom {}))
 
+(defonce pod-namespaces (volatile! {}))
```
borkdude, line 690: I specifically did not use defonce for stateful things because a lot of times I just want to clear those while testing things using `(require '[babaskha.main] :reload)`. Can we do that here too?

cap10morgan:
Hmm... not sure what is still up with the jvm tests. Latest commit got them passing on my machine. I'll try setting up my own CircleCI env to investigate.

cap10morgan:
For the linux-aarch64 failing tests, looks like I should pick a different pod or get bootleg building for arm64 too. :)

borkdude:
@cap10morgan I merged your pod PR but this caused a conflict with this PR. Can you solve it?

borkdude:
One more sync needed :)

Issue #1204, opened 2022-03-10
add property `babashka.config` to reflect `bb.edn` location


mknoszlig:
When running in a task context, both `babashka.file` and `*file*` don't contain  helpful information.
The `babashka.task` property provides a hint  that we're running in task context, but it would be nice
to have access to the location of the corresponding `bb.edn` file as well.

This could be made available on startup via a property `babaska.config`.

borkdude:
I'm ok with adding that.

PR #1206, opened 2022-03-11
add `babashka.config` property

mknoszlig:

- [x] This PR corresponds to issue [1204](https://github.com/babashka/babashka/issues/1204).

- [x] This PR contains a test to prevent against future regressions

There's something curious going on with the tests: 
```bash
> lein test :only babashka.main-test
[...]
Ran 90 tests containing 202 assertions.
0 failures, 0 errors.
```
when running all tests however:
```bash
> lein test
[...]
lein test :only babashka.main-test/config-property-test

FAIL in (config-property-test) (main_test.clj:607)
expected: (= "false\n" (test-utils/bb nil (.getPath (io/file "test-resources" "babashka" "config_property.clj"))))
  actual: (not (= "false\n" "true\ntrue\n"))
[...]
Ran 189 tests containing 563 assertions.
1 failures, 0 errors.
Tests failed.
```

I've poked at it a little bit, and it appears that the generated bb.edn file from a previous test case gets picked up, changing the behaviour. A smaller repro is to reverse the order of test clauses in `config-property-test`. `test-utils/with-config` looks good, however. i'm not sure where that binding escapes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ A preview of the next release can be installed from
 
 ## Unreleased
 
-- [#1204](https://github.com/babashka/babashka/issues/1187) add property `babashka.config` to reflect `bb.edn` location ([@mknoszlig](https://github.com/mknoszlig))
+- [#1204](https://github.com/babashka/babashka/issues/1204) add property `babashka.config` to reflect `bb.edn` location ([@mknoszlig](https://github.com/mknoszlig))
 
 ## 0.7.7 (2022-03-04)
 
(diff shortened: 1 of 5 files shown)


borkdude:
You might need to reset the property in test_utils

borkdude:
@mknoszlig here: https://github.com/babashka/babashka/blob/92ca3c6501885fd61c103bb19760c5679dae979b/test/babashka/test_utils.clj#L50

mknoszlig:
@borkdude ah, i'd put that reset in the macro, but happy to move it to the suggested position.

PR #1239, opened 2022-04-14
First pass at bash cleanup

lispyclouds:


.circleci/config.yml
```diff
@@ -11,6 +11,14 @@ commands:
           command: |
             docker run --privileged --rm tonistiigi/binfmt --install all
             docker buildx create --name ci-builder --use
+  bootstrap:
```
borkdude, line 14: Can we maybe rename the "stable" installation of bb to another name, like `bbb`? I'm afraid that we will mix up the version of bb under the build and the externally installed one and test the wrong things.

Another possibility is to invoke the bb uberjar for building bb itself using java -jar.

borkdude: I think I might even prefer the latter since then it's true bootstrapping. If bb is not stable, we should not be able to build.

lispyclouds: Yep, will rename it to `bb-released` I guess. Waiting to see if CI is happy with this.

lispyclouds: Yeah, got caught up in other things and looking at it properly now.

lispyclouds:
@cap10morgan have an eye too if you will!

.circleci/script/docker.clj
```diff
@@ -0,0 +1,92 @@
...
+
+(defn build-push-images
+  []
+  (doseq [platform (s/split platforms #",")]
+    (let [tarball-platform (s/replace platform #"\/" "-")
+          tarball-platform (if (= "linux-arm64")
+                             "linux-aarch64"
+                             tarball-platform)]
+      (fs/create-dirs platform)
+      (exec ["tar" "zxvf" (str "/tmp/release/babashka-%s-%s.tar.gz" image-tag tarball-platform) "-C" platform])
```
cap10morgan, line 71: Do you want `format` instead of `str` here?

cap10morgan (approved):
One small issue that looks like it needs addressing, but once that's done it looks great overall!

.circleci/script/docker.clj
lispyclouds: Ah yes, good spot.

lispyclouds: Also the way im injecting the labels is wrong too.

.circleci/script/docker.clj
```diff
@@ -0,0 +1,94 @@
...
+           (read-env "CIRCLE_TAG")
+           (read-env "CIRCLE_BRANCH"))
+   "--label" (str "org.opencontainers.image.version=" image-tag)])
+
+(def snapshot (= "SNAPSHOT" image-tag))
+
+(defn exec
+  [cmd]
+  (-> cmd
+      (proc/sh {:out :inherit})
```
borkdude, line 42: Probably `:inherit true` is a better choice here, to also see stderr?

Issue #1290, opened 2022-06-07
short script does not terminate workflow


borkdude:
cc @lispyclouds

lispyclouds:
Fixed in 381b9f0a4f12bc47097c4d26a9b475bcf67f0a62

borkdude:
Thanks!

PR #1294, opened 2022-06-13
use dynamic configuration to short CI

agata-anastazja:
Addresses #1290


lispyclouds:
@borkdude this is good to go now, the runners take a bit of time to provision but that can be taken care of in a followup PR. This is a good enough starting point I think.

Issue #1295, opened 2022-06-14
Programmatically starting an nrepl-server


Oddsor:
**Is your feature request related to a problem? Please describe.**
When using Babashka for long-running tasks, i.e. a simple web server, it would be nice to also be able to start an nrepl-server.

**Describe the solution you'd like**
Being able to write something like `(nrepl/start-server {:port 1234})`. The babashka.nrepl-library equivalents expect a sci-context, which hopefully can be abstracted away (🤞), but might otherwise be sufficient?

**Describe alternatives you've considered**
Chatting with @borkdude produced a nice workaround where you can start an nrepl-server using `bb nrepl-server` but also running a script-file using `--init`, for example: `bb --init webserver.clj nrepl-server <port>`

borkdude:
Looks good. PR welcome if anyone sees this, else I'll get to it in time.

PR #1297, opened 2022-06-16
Exposing babashka.nrepl.server-functions

Oddsor:
This change exposes babashka.nrepl.server-functions for programmatic use

Solves #1295 

---

### DIFF
--- a/src/babashka/main.clj
+++ b/src/babashka/main.clj
@@ -32,6 +32,7 @@
    [babashka.impl.error-handler :refer [error-handler]]
    [babashka.impl.features :as features]
    [babashka.impl.fs :refer [fs-namespace]]
+   [babashka.impl.nrepl-server :refer [nrepl-server-namespace]]
    [babashka.impl.pods :as pods]
    [babashka.impl.pprint :refer [pprint-namespace]]
    [babashka.impl.print-deps :as print-deps]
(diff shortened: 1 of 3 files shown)


Oddsor:
Barely gotten started, but figured a draft is good for discussions!

For example, I think solving this by simply exposing functions from the `babashka.nrepl.server` is sufficient, but maybe more parity with https://github.com/nrepl/nrepl/ is desirable 🤔 

Currently I don't know how to build a test for this, but after compiling and launching an nrepl-server:
```
user=> (babashka.nrepl.server/start-server! {:port 1234})
Started nREPL server at 0.0.0.0:1234
{:socket #object[java.net.ServerSocket 0x7c29cb77 "ServerSocket[addr=/0.0.0.0,localport=1234]"], :future #object[clojure.core$future_call$reify__8544 0x8d2d204 {:status :pending, :val nil}]}
```

and then connecting to it using `lein repl :connect 1234` you are greeted with:

```
$ lein repl :connect 1234
Connecting to nREPL at 127.0.0.1:1234
: Could not resolve symbol: nrepl.core/version user 
nil
Error loading namespace; falling back to user
nil
```

src/babashka/main.clj
```diff
@@ -354,6 +354,13 @@ Use bb run --help to show this help output.
        'clojure.tools.cli tools-cli-namespace
        'clojure.java.shell shell-namespace
        'babashka.core bbcore/core-namespace
+       'babashka.nrepl.server (let [ns-sci (sci/create-ns 'babashka.nrepl.server)]
+                                {'start-server! (sci/new-var 'start-server!
+                                                             (fn [opts]
+                                                               (nrepl-server/start-server! @common/ctx opts))
+                                                             {:ns ns-sci})
+                                 'stop-server! (sci/copy-var nrepl-server/stop-server! ns-sci)
+                                 'parse-opt (sci/copy-var nrepl-server/parse-opt ns-sci)})
```
borkdude, line 363: Why do we need to expose this?

borkdude: The `parse-opt` I mean

borkdude: The rest looks good, but I prefer to move this to an `babashka.impl` namespace like we do for the other namespaces.

Oddsor: Agreed, I'm not sure we need to expose parse-opt; at least I see no need for it for my uses 🤷 

I made an attempt at an impl-namespace! https://github.com/babashka/babashka/pull/1297/commits/e27ece8050d1930c1d654d2cf21d110995159ad1

Oddsor:
I've not written a test for this, mostly because I'm not sure if it's necessary? 🤔 
There are separate nrepl-tests, and a "publics-namespace-test" that already covers most of the additions.

I've manually tested and found it working; a minor bug I found is that stopping the server doesn't kill existing connections (just seems to stop listening for new ones), so maybe a separate issue can be filed for that!

PR #1301, opened 2022-06-21
Optimise executors when shorted

lispyclouds:


lispyclouds:
@borkdude good to go from my side, tested with a shorting <5s on an average.

PR #1362, opened 2022-09-13
Add and expose babashka.classes/all-classes

eerohele:

- [x] This PR corresponds to an [issue with a clear problem statement](https://github.com/babashka/babashka/blob/master/doc/dev.md#start-with-an-issue-before-writing-code):

   To offer editor auto-completion for the Java classes Babashka supports, we need to expose a list of those classes to users.

   Question: should `all-classes` return a list of `java.lang.Class` instances (as in this PR), or a list of symbols?

- [ ] This PR does not contain a [test](https://github.com/babashka/babashka/blob/master/doc/dev.md#tests) to prevent against future regressions.

   Maybe it should? Any thoughts on how you'd like to have this tested? Testing is somewhat complicated by the fact that not all class names (e.g. `[Ljava.time.temporal.TemporalField;`) are read-able.

### DIFF
--- a/src/babashka/impl/classes.clj
+++ b/src/babashka/impl/classes.clj
@@ -5,6 +5,7 @@
    [babashka.impl.proxy :as proxy]
    [cheshire.core :as json]
    [clojure.core.async]
+   [sci.core :as sci]
    [sci.impl.types :as t]))
 
 (def base-custom-map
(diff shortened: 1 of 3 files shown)


borkdude:
Maybe add a simple test like `(every? class? (babashka.classes/all-classes))` which returns true

eerohele:
Done.

PR #1372, opened 2022-09-22
Add single arg read method to PipedInputStream proxy

retrogradeorbit:


retrogradeorbit:
OK, bear with me. This will require some explanation.

I have been implementing `babashka.process/process` piped stream support in the upcoming pod `bbssh`. During this work the following code would not operate correctly: https://github.com/epiccastle/bbssh/blob/main/test/bb_test/test_password_exec.clj#L247-L251

That is, I could stream from `babashka.process/process` to `bbssh/exec` ok, but not the other way around, from `bbssh/exec` to `babashka.process/process`.

It turns out that `PipedInputStream` objects do not actually have a single argument (2 arity) `.read` method. https://docs.oracle.com/javase/7/docs/api/java/io/PipedInputStream.html#read() . (This is why I originally didn't include such a method signature in the original proxy implementation code I submitted to babashka)

And `clojure.java.io/copy` (used to connect the input stream to the output stream in babashka.process), when copying from input stream to output stream, calls the single argument (2 arity) version of `.read` on the input stream https://github.com/clojure/clojure/blob/master/src/clj/clojure/java/io.clj#L305

Thus, the io/copy used by process fails raiusing an exception.It is inside a `with` block, so the output stream (input) is closed. And its inside a future so you dont see the exception. The closed input stream means the process sees no input and proceeds as if stdin was closed.

This can be remedied by building a custom proxy that includes this version of `.read`. It won't alter PipedInputStream objects coming from other packages (JSch) but we can build a proxy in front of them that implements this extra method signature and provides an implementation (which is what I have done in bbssh to get process streaming working).

I didn't include a test as the change is fairly declarative and so small. I can open an issue if needed but thought it's such a small change.

retrogradeorbit:
In clojure you cannot actually io/copy from a PipedInputStream to an OutputStream (but you can to other destinations because that uses a different io/copy multimethod). In clojure you **can** create a proxy with a 2 arity read. 

```Clojure 1.11.1
user=> (def p (proxy [java.io.PipedInputStream] [] (read [a] (println "HIT") -1)))
#'user/p
user=> (.read p nil)
HIT
-1
```

I think there is a combination of things at play here. One is what I would consider a mistake way back in the initial implementation in Java 7(?) of PipedInputStream. There should have been an implementation of this method (it is defined in the java.io.InputStream interface). Two is I think that io/copy should probably use the 4 arity read call instead, because of this original java omission.

Taking an instance where a third party library returns a PipedInputStream. In clojure we could fill in the missing read implementation by using proxy, passing every call straight through to the underlying PipedInputStream instance, and calling the 4 arity read implementation from the new 2 arity. This would create a proxy that could then be used with io/copy.

We should probably be able to do the same in bb.

This is the current version of bb:

```
user=> (def p (proxy [java.io.PipedInputStream] [] (read [a] (println "HIT") -1)))
#'user/p
user=> (.read p nil)
Wrong number of args (2) passed to: babashka.impl.proxy/proxy-fn/fn--21688
```

and this is with the patch:

```
user=> (def p (proxy [java.io.PipedInputStream] [] (read [a] (println "HIT") -1)))
#'user/p
user=> (.read p nil)
HIT
-1
```

PR #1374, opened 2022-09-24
Force as global opt

### LINKED ISSUE #1336: tasks subcommand doesn't work with global --force option
**version**

v0.8.156

**platform**

Linux (Nix)

**problem**

Specifying the global `--force` option before the `tasks` subcommand results in Babashka trying to execute a file called `tasks` rather than the `tasks` subcommand. Running `bb tasks --force` instead has the desired behaviour.

**repro**

```
$ bb --force tasks
Type:     java.lang.Exception
Message:  File does not exist: tasks
```

**expected behavior**

The `tasks` subcommand runs with forced classpath recomputation.

bobisageek:
Closes #1336 

This change moves `--force` from opt parsing to global opt parsing.

I was initially worried about regressions from removing force as an option, but based on some cursory arg parsing testing on master, it looked like (in addition to the issue reported), `--force` often either doesn't set `:force?`, or causes commands that come after it to be treated as a file name (the reported issue is one case of this).

- [x] This PR corresponds to #1336 (https://github.com/babashka/babashka/blob/master/doc/dev.md#start-with-an-issue-before-writing-code).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ A preview of the next release can be installed from
 
 ## Unreleased
 
+- [#1336](https://github.com/babashka/babashka/issues/1336): tasks subcommand doesn't work with global --force option ([@bobisageek](https://github.com/bobisageek))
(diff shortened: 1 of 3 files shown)


borkdude:
Thanks!

PR #1378, opened 2022-09-28
Mark bb.edn load exception

lispyclouds:
This improves the error visibilty when loading a `bb.edn` by marking the it better instead of just the stacktrace which could be confusing.

### DIFF
--- a/src/babashka/main.clj
+++ b/src/babashka/main.clj
@@ -1049,7 +1049,7 @@ Use bb run --help to show this help output.
        (catch java.lang.RuntimeException e
          (if (re-find #"No dispatch macro for: \"" (.getMessage e))
            (throw (ex-info "Invalid regex literal found in EDN config, use re-pattern instead" {}))
-           (do (println "Error during loading bb.edn:")
+           (do (.println *err* "Error during loading bb.edn:")
                (throw e))))))
 
 (defn main [& args]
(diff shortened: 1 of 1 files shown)


src/babashka/main.clj
```diff
@@ -1049,7 +1049,8 @@ Use bb run --help to show this help output.
        (catch java.lang.RuntimeException e
          (if (re-find #"No dispatch macro for: \"" (.getMessage e))
            (throw (ex-info "Invalid regex literal found in EDN config, use re-pattern instead" {}))
-           (throw e)))))
+           (do (println "Error during loading bb.edn:")
```
borkdude, line 1052: can you print to stderr instead?

src/babashka/main.clj
```diff
@@ -1049,7 +1049,8 @@ Use bb run --help to show this help output.
        (catch java.lang.RuntimeException e
          (if (re-find #"No dispatch macro for: \"" (.getMessage e))
            (throw (ex-info "Invalid regex literal found in EDN config, use re-pattern instead" {}))
-           (throw e)))))
+           (do (.println *err* "Error during loading bb.edn:")
```
borkdude, line 1052: I think this will result in reflection.

Can you do `(binding [*out* *err*] (println ...))`?

lispyclouds: pushed

PR #1406, opened 2022-10-24
Add fs lib test

### LINKED ISSUE #602: Add tests for raynes/fs lib and add to libraries.md
This works:

``` 
export BABASHKA_CLASSPATH=$(clojure -Spath -Sdeps '{:deps {clj-commons/fs {:mvn/version "RELEASE"}}}')
$ clojure -M:bb-local -e "(require '[me.raynes.fs :as fs]) (fs/glob \"*.md\")"
(#object[java.io.File 0x78f2bb1b "/Users/borkdude/Dropbox/dev/clojure/babashka/CHANGELOG.md"] #object[java.io.File 0x651c826c "/Users/borkdude/Dropbox/dev/clojure/babashka/README.md"] #object[java.io.File 0x10323056 "/Users/borkdude/Dropbox/dev/clojure/babashka/CONTRIBUTING.md"])
``` 

but it would be nice if we added some tests to lib-tests.

Also we need to add it to the libraries.md page.

- [x] Fix #601 
- [ ] A good test would be to check if the Java 7 conditional is executed:

``` 
$ clojure -A:bb-local -cp $(lein classpath):src:test -e "(require '[me.raynes.fs :as fs]) (contains? (ns-publics 'me.raynes.fs) 'directory?)"
``` 

- [ ] Test if the Coercions impl is working

bobisageek:

Closes #602.
The library tests use midje - I did a little experimenting to see how compatible midje itself would be, but based on that experimenting, it'd need some extra work, so I mocked the `fact` macro, which is enough for a big subset of the tests.

I just commented the tests that need other midje things to make as little change to the test file as necessary (and because we could maybe mock more of midje or get midje working in the future, and then more of the tests could be uncommented). However, if preferred, I can go in and rip out the unsupported tests (all of them, or even just the ones that use the compression namespace for which bb is missing some classes) to make the file less noisy.

Outside of that, this change is basically just the result of using the add libtest script to add the fs lib.


borkdude:
@bobisageek 

Thanks! 

Recently I found a way not to copy all the code from the tests but to clone it using a git dep and then add the test directories to the classpath. This saves adding a couple of things to bb's git repo. For examples of this, check out the latest entries in the tested-libs.edn file. 
It might not be possible to do this for all namespaces / resources (the ones with BB-TEST-PATCH), but might be possible for some. Have you checked that?

What was the matter with midje?

bobisageek:
re: the git dep thing: I haven't used that approach yet, but it looks nice and straightforward. In this case, the test ns is heavily patched, including a separate ns for mocking out the macro from midje.

re: midje, there are a few classes it needs, but it also has some deftypes. I didn't go too deep into it - I wanted to avoid conflating the lib tests and midje compatibility. 
Dependent on size inflation and what not, would you be interested in midje compatibility? I can dig deeper into it, and see what it'd take to get that working (at which point there'd be a lot less patching for these tests).

PR #1450, opened 2022-12-13
fix #1280

### LINKED ISSUE #1280: Babashka REPL doesn't quit right after ctrl-d
It quits after the second ctrl-d or ctrl-d followed by an enter though.

formerly-a-trickster:
the problem is that parser/parse-next waits for an enter before returning an object we can use peek-char to pre-emptively react to any EOF at the start of the line in case of not-start-of-line, parse-next takes over

- [x] This PR corresponds to an [issue with a clear problem statement](https://github.com/babashka/babashka/blob/master/doc/dev.md#start-with-an-issue-before-writing-code), issue #1280


borkdude:
Thanks a lot!

Issue #1473, opened 2023-01-24
Make relative paths in deps of bb.edn resolve relative to the bb.edn


lispyclouds:
**Is your feature request related to a problem? Please describe.**
Currently, due to its nature of implementation in [deps.clj](https://github.com/borkdude/deps.clj) and its use in bb, if relative paths via `:local/root` are used in the `:deps` of bb.edn, they are resolved relative to the pwd of the invocation and not the bb.edn.

**Describe the solution you'd like**
To make the expectation of all the paths in bb.edn resolve relative to it consistent the resolution of these paths should be changed as well.

**Describe alternatives you've considered**
None.

**Additional context**
Some points to consider:
- Since this is shelling out to the usual deps resolver, how does the classpath caching get affected by this change? Would it pollute the pwd now?

borkdude:
The only change necessary to fix this is to use `(:bb-deps-root ...)` and bind the `deps.clj` `*dir*` variable to it (only if it's set, `with-bindings` allows you to do that conditionally.

PR #1478, opened 2023-01-27
[#1473] Make deps resolve paths relative to bb.edn

lispyclouds:
- [ ] tests
- [x] tests for deps-root
- [x] changelog

### DIFF
--- a/src/babashka/impl/deps.clj
+++ b/src/babashka/impl/deps.clj
@@ -89,10 +89,11 @@
                             "-Sdeps-file" "") ;; we reset deps file so the local deps.edn isn't used
                  args (if force (cons "-Sforce" args) args)
                  args (concat args [(str "-A:" (str/join ":" (cons ":org.babashka/defaults" aliases)))])
+                 deps-root (:deps-root @bb-edn)
                  bindings (cond->
                             {#'deps/*env* env
                              #'deps/*extra-env* extra-env}
-                            (:deps-root @bb-edn) (assoc #'deps/*dir* (:deps-root @bb-edn)))
+                            deps-root (assoc #'deps/*dir* deps-root))
                  cp (with-out-str (with-bindings bindings
                                     (apply deps/-main args)))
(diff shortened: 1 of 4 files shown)


lispyclouds:
Has the same flaky failures, but seems to be okay.

borkdude:
It would be nice to have a test for this. Filling out the PR template would be nice too which reminds you of this.

src/babashka/impl/deps.clj
```diff
@@ -89,8 +89,12 @@
                             "-Sdeps-file" "") ;; we reset deps file so the local deps.edn isn't used
                  args (if force (cons "-Sforce" args) args)
                  args (concat args [(str "-A:" (str/join ":" (cons ":org.babashka/defaults" aliases)))])
-                 cp (with-out-str (binding [deps/*env* env
-                                            deps/*extra-env* extra-env]
+                 deps-root (:deps-root @bb-edn)
```
borkdude, line 92: What I meant was that deps-root is used higher up, so pulling out some levels higher lets us re-use it.

lispyclouds:
@borkdude if i add a tests for the relative to bb.edn with something like
```clojure
(spit config
      '{:paths ["src"]
         :deps  {local/dep {:local/root "local-dep"}}
         :tasks {cp (prn (babashka.classpath/get-classpath))}})
(testing "relative paths in deps should be relative to bb.edn"
  (let [root (fs/create-dir (fs/file dir "local-dep"))
        _ (spit (str (fs/file root "deps.edn")) {})
        out     (bb "--config" config "cp")
        entries (cp/split-classpath out)]
    (is (= (fs/parent f) (fs/parent (first entries))))))
```
this fails to create the cache folder with (ran on Mac):
```
==== Testing JVM version

lein test babashka.bb-edn-test
=== deps-root-tests

Error building classpath. Can't create directory: /private/Users/rahul.de/.clojure/.cpcache
Tests failed.
```

Any suggestions?

PR #1483, opened 2023-02-04
add lib tests for xforms

### LINKED ISSUE #1000: Add xforms tests to CI (was: make xforms run with bb)
**Is your feature request related to a problem? Please describe.**
https://github.com/cgrand/xforms is a library of extra transducers, reducing fns, and transducing contexts, which greatly improves the quality of life when someone tries to maximize the use of transducers in clojure code.

**Describe the solution you'd like**
I would love to be able to use the `xforms` library in Babashka, extra points if it's available out of the box.

**Describe alternatives you've considered**
One alternative is to not use transducers in situations where it is clunky with what's available in clojure.core. This comes with unnecessary laziness and collection allocations but I guess that only becomes a bottleneck in heavily performance-sensitive code or code that runs a lot of times.

Another is to write those utility functions myself in a bb-compatible way, but there probably is a reason why xforms is not trivial code.

This isn't an urgent request by any means, and it is quite likely that the xforms library itself will need to be updated to support bb properly. One such piece of code is a [multi-class reify](https://github.com/cgrand/xforms/blob/62375212a8604daad631c9024e9dbe1db4ec276b/src/net/cgrand/xforms.cljc#L101). The lib has not recently been updated although there are a few open PRs, including one that fixes a handful of reflection warnings.

bobisageek:
Closes #1000 

Tests work with no patches needed, so this is just a "pull the tests from the git lib and execute them" addition.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ A preview of the next release can be installed from
 
 [Babashka](https://github.com/babashka/babashka): Native, fast starting Clojure interpreter for scripting
 
+## Unreleased
+- [#1000](https://github.com/babashka/babashka/issues/1000): add lib tests for xforms ([@bobisageek](https://github.com/bobisageek))
+
 ## 1.1.173 (2023-02-04)
 
 - [#1473](https://github.com/babashka/babashka/issues/1473): support `--config` in other dir + `:local/root` ([@lispyclouds](https://github.com/lispyclouds))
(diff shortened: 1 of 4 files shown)


borkdude:
Hey @bobisageek - thanks a lot!

For the failing tests: I've added `^:flaky` metadata on some of those tests and I think I want to disable those in CI when on the master branch. 

I haven't looked deeply into that yet, but if you have any ideas, feel free to poke around with that as well.
Another way to go about it is to remove the `:test` metadata from the vars that have flaky metadata in the test namespaces, so we don't have to do any Windows vs *Nix scripting.

PR #1486, opened 2023-02-05
skip flaky tests on master on AppVeyor

bobisageek:
filed as a draft for now, in case you'd like to test with a different base branch first...

- update `test.bat` to pass args along to the `lein test` call, similar to the bash script
- add `:excludes` test selector based roughly on test-runner
- appveyor yaml changes:
  - move test steps from build to test phase
  - test prep steps go to `before_test` (mainly to exclude them from the conditionals)
  - default test script (on branches other than master/no PR info) calls test.bat with windows selector
  - master branch with no PR info calls test.bat excluding both the "skip on windows" and the "flaky" selectors

The main concern I have is that the environment (i.e. BABASHKA_TEST_ENV) might not be maintained from `before_test` to `test_script`, but the appveyor docs make it sound like it should.

### DIFF
--- a/.build/bb.edn
+++ b/.build/bb.edn
@@ -3,4 +3,6 @@
         #_{:local/root "../gh-release-artifact"}
         {:git/url "https://github.com/borkdude/gh-release-artifact"
          :git/sha "ce060c12a25b552b864dc90f8fb344a2eb91ea9d"}}
- :tasks {release-artifact babashka.release-artifact/release}}
+ :tasks {release-artifact babashka.release-artifact/release
+         test             {:requires ([babashka.release-artifact])
+                           :task (apply shell (babashka.release-artifact/bb-tests-cmd))}}}
--- a/.build/bb.edn
+++ b/.build/bb.edn
@@ -3,6 +3,4 @@
         #_{:local/root "../gh-release-artifact"}
         {:git/url "https://github.com/borkdude/gh-release-artifact"
          :git/sha "ce060c12a25b552b864dc90f8fb344a2eb91ea9d"}}
- :tasks {release-artifact babashka.release-artifact/release
-         test             {:requires ([babashka.release-artifact])
-                           :task (apply shell (babashka.release-artifact/bb-tests-cmd))}}}
+ :tasks {release-artifact babashka.release-artifact/release}}
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -75,24 +75,9 @@ before_test:
 
     set BABASHKA_TEST_ENV=native
     
-for:
-- branches:
-    only: 
-      - master
-  test_script:
-    # if master and PR# is not present, it should be a build of master, so skip flaky tests
-    # else it's a PR to master, so don't skip flaky tests
-    - ps: >-
-        if(-not $env:APPVEYOR_PULL_REQUEST_NUMBER) {
-        script/test.bat :excludes "[:skip-windows :flaky]"
-        } else {
-        script/test.bat :windows
-        }
-        script/run_lib_tests.bat
-
 test_script:
 - cmd: >-
-    call script/test.bat :windows
+    bb --config .build/bb.edn --deps-root . run test
 
     call script/run_lib_tests.bat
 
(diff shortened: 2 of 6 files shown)


borkdude:
Looks good!

> The main concern I have is that the environment (i.e. BABASHKA_TEST_ENV) might not be maintained from before_test to test_script, but the appveyor docs make it sound like it should.

The test env is printed once the tests are running.

This PR only sets it up for Windows. What about the others?

bobisageek:
Well, I wanted to take it one step at a time... I figured I could work out any little kinks with appveyor, and then dive into the circle config, so it's not a case where like three separate things break. And I'm glad I did... I _think_ AppVeyor might be failing the test because something printed to stderr (based on the output basically echoing the warning message), so I need to dig into that.

borkdude:
Programmatically disabling the tests from within the test namespaces based on the branch name and perhaps the environment variable `CI=true` (which is usually set) is also still an option. It seems to be a less CI-specific change and once we move things over from AppVeyor to something else, I'd not have to re-implement specific steps.

bobisageek:
So, after a bit of experimenting, it seems like specifically when you run things with powershell on appveyor, outputting to stderr causes a failure (if you use `cmd`, it doesn't care). So much for better of two evils...

Regarding disabling the tests, the messy part is deciding whether or not it's a PR (on AppVeyor). If you're ok with not running flaky tests on PRs to master, then (I think) we can just do the branch thing.

borkdude:
Can't we just do: if it's a PR, always run the flaky tests?

borkdude:
I would also be ok with _always_ running flaky tests but not letting the build fail, e.g. just print a warning or so.
It's pretty annoying to have PRs fail as well due to flaky tests sometimes.

bobisageek:
The messy part is deciding if it's a PR without batch scripting (in the context of the appveyor file). It's sort of a catch 22; using CMD means ugly scripting for determining PR-ness, but using PS means that any writes to stderr from the test scripts cause a failure (I was able to verify this on a simple repro project in AppVeyor).

I'm thinking about mostly abandoning this approach and throwing an additional task into .build/bb.edn to use bb to replace the batch/powershell stuff, which would allow re-use of the current_branch function and lets us move away from any batch (or even OS-specific) scripting, and would also mean not having different customizations in AppVeyor's yaml vs CircleCI yaml vs....

To some extent, I think there will still be some CI-provider-specific stuff, but it would hopefully just be environment variables (the same thing current_branch does, but aimed at "is this a PR?"). I'm hoping to play around with this some more after work.

borkdude:
@bobisageek Why is deciding PR hard? There's an environment variable set for the PR number in appveyor?

But yes, I think I prefer the bb scripting approach too.

bobisageek:
I think I have a bb task that'll work on linux and windows and with the 4 current CI platforms (I still want to get AppVeyor working first before trying to expand it out). I'll push it after I get some sleep and can watch the build finish.

bobisageek:
With the latest push, there's a new fn `bb-tests-cmd` in the release-artifact ns (I should maybe move that to a different ns, but I'd want to require that ns anyway to get the `current-branch` fun). That fun builds a vector of the command and args to call the test script with the appropriate exclusions based on OS and branch/PR. Then there's a new task in `.build/bb.edn` that calls `apply shell` on those args.

On the plus side, generating the args means there's one place to deal with for skipping the flaky tests. Looking over the other CIs, Circle should be ok.. we can generate the command line and inject that into the generated CI script. Github actions is maybe a somewhat gray area - test gets called early; we can use `lein bb` to run the task, but it's sort of a chicken-egg situation in that we're using the bb code to run the task that will test the bb code.

It looks like AppVeyor is passing now - any changes you'd like to see in the bb scripting before I start on tweaking the Circle/GH Actions builds?

P.S. - apologies if I'm overcomplicating things

bobisageek:
That is probably easier to deal with... I'll make the selectors that are in use exclude flaky and add a flaky run to the test script that always succeeds. Like you said, we won't catch failures that aren't related to flakiness.

bobisageek:
... and of course, when you _want_ the flaky tests to fail, they all pass

borkdude:
@bobisageek Thanks both for your persistence, time and effort!

PR #1503, opened 2023-02-25
Prevent uberscript/uberjar overwrites

### LINKED ISSUE #1489: Mixed source and dest file using uberscripts blanks files
I know did not use the tool correct, but feel is dangerus it wipes a sourcefile!

**version**
babashka v1.1.173

**platform**
Windows 11

**problem**
Tested the Unberscript and mixed source and dest file witch did blank my sourcefile. 
, mabe should give warning that a file with data exits ?

```
Directory: C:\test\bb_test

Filename: 
-a----        2023-02-09     14:32             70 source.clj

bb uberscript .\source.clj

Result:

-a----        2023-02-09     15:11              0 source.clj

```

bobisageek:
closes #1489 

Summary of changes:
- `file-write-allowed?` checks that the file is a jar or empty/non-existent
- before writing uberjar or uberscript, check that file write is permitted (for uberjar specifically, I tweaked the classpath check into a cond instead of nesting ifs)
- add tests that check:
  - uberscript doesn't overwrite non-empty target
  - uberjar does overwrite non-empty jar
  - uberjar doesn't overwrite non-empty file other than jar (simulating a source file being specified)

We can definitely change the 'overwrite prohibited' message if there's different wording you'd prefer.

### DIFF
--- a/src/babashka/main.clj
+++ b/src/babashka/main.clj
@@ -780,6 +780,14 @@ Use bb run --help to show this help output.
       env-os-name-present? (not= env-os-name sys-os-name)
       env-os-arch-present? (not= env-os-arch sys-os-arch))))
 
+(defn file-write-allowed?
+  "For output file of uberscript/uberjar, allow writing of jar files
+   and files that are empty/don't exist."
+  [path]
+  (or (= "jar" (fs/extension path))
+    (not (fs/exists? path))
+    (zero? (fs/size path))))
+
 (def seen-urls (atom nil))
 
 (defn read-data-readers [url]
@@ -1062,12 +1070,15 @@ Use bb run --help to show this help output.
                 1)]
         (flush)
         (when uberscript
-          (let [uberscript-out uberscript]
-            (spit uberscript-out "") ;; reset file
-            (doseq [s (distinct @uberscript-sources)]
-              (spit uberscript-out s :append true))
-            (spit uberscript-out preloads :append true)
-            (spit uberscript-out expression :append true)))
+          (if (file-write-allowed? uberscript)
+            (do
+              (spit uberscript "")                        ;; reset file
+              (doseq [s (distinct @uberscript-sources)]
+                (spit uberscript s :append true))
+              (spit uberscript preloads :append true)
+              (spit uberscript expression :append true))
+            (throw (Exception. (str "Uberscript target file '" uberscript
+                                 "' exists and is not empty. Overwrite prohibited.")))))
         (when uberjar
           (if-let [cp (cp/get-classpath)]
             (let [uber-params {:dest uberjar
--- a/src/babashka/main.clj
+++ b/src/babashka/main.clj
@@ -1080,22 +1080,28 @@ Use bb run --help to show this help output.
             (throw (Exception. (str "Uberscript target file '" uberscript
                                  "' exists and is not empty. Overwrite prohibited.")))))
         (when uberjar
-          (if-let [cp (cp/get-classpath)]
-            (let [uber-params {:dest uberjar
-                               :jar :uber
-                               :classpath cp
-                               :main-class main
-                               :verbose debug}]
-              (if-let [bb-edn-pods (:pods @common/bb-edn)]
-                (fs/with-temp-dir [bb-edn-dir {}]
-                  (let [bb-edn-resource (fs/file bb-edn-dir "META-INF" "bb.edn")]
-                    (fs/create-dirs (fs/parent bb-edn-resource))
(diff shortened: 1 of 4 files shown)


src/babashka/main.clj
```diff
@@ -780,6 +780,14 @@ Use bb run --help to show this help output.
       env-os-name-present? (not= env-os-name sys-os-name)
       env-os-arch-present? (not= env-os-arch sys-os-arch))))
 
+(defn file-write-allowed?
```
borkdude, line 783: Maybe we could even just allow `.jar` files, or would this be too restrictive? I don't see a reason why you would want to write a `.txt` file. Bb itself also relies on the `.jar` file extension being there when executing uberjars.

test/babashka/main_test.clj
```diff
@@ -487,6 +487,14 @@
     (is (empty? (bb nil "--uberscript" (test-utils/escape-file-paths (.getPath tmp-file)) "-e" "(System/exit 1)")))
     (is (= "(System/exit 1)" (slurp tmp-file)))))
 
+(deftest uberscript-overwrite-test
```
borkdude, line 490: I'm not sure why there are tests for this in two different places, one of them may be enough?

bobisageek: I agree. I put the new uberjar tests in uberjar because I look on it as a 'sibling' of the empty classpath check, so what I'd like to do is pull both `throw-on-empty-classpath` and `target-file-overwrite-test` from uberjar-test into main-test

src/babashka/main.clj
bobisageek: Well, there's a use case I could see that I didn't want to risk breaking in case it's out there in the wild:
On GUI systems where I can associate extensions with an app/command line, maybe I want to associate a separate-but-related extension with bb, so `.jar` would run with `java -jar` by default, but maybe `.bar` or `.bbjar` would run with `bb -jar`.
I actually personally had a use case for something just like this, but I ended up making a separate executable (because I used instaparse).
For that scenario, I might want to use a `.bar` (or similar) extension with uberjar.

That's my case for it potentially being too restrictive, but maybe that's a bit too contrived.

borkdude: Yeah, that's a good reason to not restrict it to `.jar`.

There is now also https://github.com/babashka/instaparse-bb

bobisageek:
I think the test failure is essentially a flaky test - nghttp2 seems to have gone unavailable in the time window, but I'm looking to see if I can (at least anecdotally) verify that locally.

bobisageek:
yeah, lib tests now fail/hang at hato.client-test on linux for me

it seems we could use http-2 with httpbin to make the flaky/timeout thing work with `test-http2` as well... a cursory preliminary test suggests that httpbin now uses http-2 as default:

```clj
user=> (require '[hato.client :refer [get]])
nil
user=> (:version (get "https://httpbin.org/anything"))
:http-2
user=> (:version (get "https://httpbin.org/anything" {:version :http-2}))
:http-2
user=> (:version (get "https://httpbin.org/anything" {:version :http-1.1}))
:http-1.1
```

src/babashka/main.clj
```diff
@@ -780,6 +780,14 @@ Use bb run --help to show this help output.
...
+  "For output file of uberscript/uberjar, allow writing of jar files
+   and files that are empty/don't exist."
+  [path]
+  (or (= "jar" (fs/extension path))
+    (not (fs/exists? path))
+    (zero? (fs/size path))))
```
borkdude, line 789: Why would we overwrite an empty existing file?

bobisageek: I'd say philosophically, because there's no content to prevent the loss of. Practically, this check serves to accommodate the tests that create temp files for the outputs.

borkdude: I still think it's a little weird that when we would e.g. have an empy `foo.clj` file, which is valid Clojure, we would overwrite it with a jar file.

> Practically, this check serves to accommodate the tests that create temp files for the outputs.

Can't the test delete the file before the real test?

bobisageek: > Can't the test delete the file before the real test?

yeah. I wasn't sure if the value prop was there, but I've removed the zero size allowance and updated the uberscript tests; re-running my local compile/test battery now

borkdude:
@bobisageek Thanks! About hato tests, improvements welcome of course

bobisageek:
Removing the zero size allowance didn't get pushed (my linux build runs slow in WSL). I'll wrap that up and include it with updates to the hato test

borkdude:
Ah, sorry... :)

Issue #1521, opened 2023-03-15
Also push to GHCR apart from Dockerhub


lispyclouds:
**Is your feature request related to a problem? Please describe.**
Docker is being weird and we'd like to move to GHCR.

**Describe the solution you'd like**
We switch to pushing to GHCR for now.

**Describe alternatives you've considered**
https://github.com/Quantisan/docker-clojure/issues/129

**Additional context**
Slack conversation: https://clojurians.slack.com/archives/CLX41ASCS/p1678813448403629

PR #1522, opened 2023-03-15
[#1521] Push images to GHCR

lispyclouds:

### DIFF
--- a/.circleci/script/docker.clj
+++ b/.circleci/script/docker.clj
@@ -46,6 +46,10 @@
 
 (defn docker-login
   [username password]
+  (exec ["docker" "login" "-u" username "-p" password]))
+
+(defn docker-login-ghcr
+  [username password]
   (exec ["docker" "login" "ghcr.io" "-u" username "-p" password]))
 
 ;; TODO: Remove this when Dockerhub goes off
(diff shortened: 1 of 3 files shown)


lispyclouds:
I have added those env var in circle using a classic token i generated with packages permissions. couldnt figure out the permissions for pacakges on the new tokens. the user is `lispyclouds` for now.

borkdude:
@lispyclouds I think we can keep publishing to dockerhub until they shut things off. ghcr would be an additional thing people can move to over time.

lispyclouds:
okay, will just do it as an extra tag and push

borkdude:
LGTM. Shall we test this before merge to master, by removing the master only check temporarily?

lispyclouds:
doing the build twice due to:
- the build and push are complected due to the multiarch buildx things
- its easier to simply remove the dockerhub code when moving than having complicated branched code

lispyclouds:
merging to master should be safe enough, this is an accretive change for me.

lispyclouds:
should be good now

borkdude:
"this is an accretive change": I see enough points that could go wrong even with a small change.

does logging in twice undo the first login, or can you do multiple logins at the same time?

borkdude:
and the thing I wanted to test mostly is if the github containers are really pushed: how is that evident? I've had credential-related failures enough that it warrants testing on this branch before going to master.

lispyclouds:
makes sense. taking off the PR and the master check for the test.

> does logging in twice undo the first login, or can you do multiple logins at the same time?

multiple logins can exist as long as their URLs are different, which is the case here.

borkdude:
And now? :)

lispyclouds:
well want expecting this to pass in the first go but more from _can debug on master with less code changes_ 😛

borkdude:
yeah right ;)

PR #1523, opened 2023-03-21
fix(docker): make image smaller

raszi:
By removing the package list we could reduce the Docker image size even further from 217MB to 174MB.


borkdude:
@lispyclouds wanna take this one?

lispyclouds (approved):

PR #1539, opened 2023-04-17
vary ports across socket repl tests

bobisageek:

This change uses different ports for each socket repl test, which seems (so far) to eliminate intermittent connection exceptions. To facilitate this, the `socket-command` function is changed  to `socket-command-on-port`, which takes a port and then returns the existing `socket-command` function (with the port closed over). Each test then `let`s its own `socket-command` with the different ports.


borkdude:
Thanks!

PR #1558, opened 2023-05-13
Support babashka.process reload for exec fn

### LINKED ISSUE #1557: Unable to reload `babashka.process/exec`
**version**
```
❯ bb --version
babashka v1.3.179
```

**platform**
Repro on Linux, but should apply to all.

**problem**
Babashka comes bundled with `babashka.processs` built in.
Sometimes a newer version of `babashka.process` becomes available.

For example, at the time of this writing, a newer version of `babashka.process` is available with a `:cmd` `opt` feature.

A user can opt to use the newer version via a ns reload.
This works great in general, but not for the `exec` function.

**repro**
Given a `bb.edn` that specifies the new version:
```
{:deps {babashka/process {:mvn/version "0.5.19"}}}
```

Here's a REPL session that shows the behaviour.
We'll start off using built-in babashka.process:
```Clojure
❯ rlwrap bb
Babashka v1.3.179 REPL.
Clojure rocks, Bash reaches.

user=> (require '[babashka.process :as p])
nil
```
And now let's call a new feature that is not yet in bb (it will fail):
```Clojure
user=> (p/shell {:cmd ["ls" "-l"]})
java.lang.NullPointerException [at <repl>:2:1]
```
And we'll reload the ns to get the updated version specified in `bb.edn` and retry our command (it will work)
```Clojure
user=> (require '[babashka.process :as p] :reload)
nil
user=> (p/shell {:cmd ["ls" "-l"]})
total 4
-rw-rw-r-- 1 lee lee 51 May 13 12:39 bb.edn
{:proc #object[java.lang.ProcessImpl 0x6d1bac37 "Process[pid=66629, exitValue=0]"], :exit 0, :in #object[java.lang.ProcessBuilder$NullOutputStream 0x73925281 "java.lang.ProcessBuilder$NullOutputStream@73925281"], :out #object[java.lang.ProcessBuilder$NullInputStream 0x1affa17d "java.lang.ProcessBuilder$NullInputStream@1affa17d"], :err #object[java.lang.ProcessBuilder$NullInputStream 0x1affa17d "java.lang.ProcessBuilder$NullInputStream@1affa17d"], :prev nil, :cmd ["ls" "-l"]}
```

Now let's try a call to `exec` to show the issue:
```
user=> (p/exec "ls -l")
clojure.lang.ExceptionInfo: exec is not supported in non-GraalVM environments [at babashka/process.cljc:612:8]
```

**expected behavior**
`exec` should also work after ns reload.

**thoughts**
I am guessing that if bb exposed `org.graalvm.nativeimage.ProcessProperties/exec` this should work.

**next steps**
I'll experiment and follow up.

lread:
Exposed GraalVM `ProcessProperties/exec` signature used by babashka.process/exec.

Add new `graal?` feature (on by default) to allow folks to build/use babashka without this specific Graal API.

On my linux dev box bb executable increased by 8kb.

Closes #1557

### DIFF
--- a/src/babashka/impl/classes.clj
+++ b/src/babashka/impl/classes.clj
@@ -15,6 +15,9 @@
 (def has-domain-sockets?
   (resolve 'java.net.UnixDomainSocketAddress))
 
+(def has-graal-process-properties?
+  (resolve 'org.graalvm.nativeimage.ProcessProperties))
+
 (def base-custom-map
   `{clojure.lang.LineNumberingPushbackReader {:allPublicConstructors true
                                               :allPublicMethods true}
@@ -183,7 +186,8 @@
                             {:methods [{:name "getBundle"
                                         :parameterTypes ["java.lang.String","java.util.Locale",
                                                          "java.lang.ClassLoader"]}]})
-    (resolve 'org.graalvm.nativeimage.ProcessProperties)
+
+    has-graal-process-properties?
     (assoc `org.graalvm.nativeimage.ProcessProperties
            {:methods [{:name "exec"}]})))
 
--- a/src/babashka/impl/classes.clj
+++ b/src/babashka/impl/classes.clj
@@ -182,7 +182,12 @@
                             `java.util.ResourceBundle
                             {:methods [{:name "getBundle"
                                         :parameterTypes ["java.lang.String","java.util.Locale",
-                                                         "java.lang.ClassLoader"]}]})))
+                                                         "java.lang.ClassLoader"]}]})
+    features/graal? (assoc `org.graalvm.nativeimage.ProcessProperties
+                           {:methods [{:name "exec"
+                                       :parameterTypes ["java.nio.file.Path"
+                                                        "[Ljava.lang.String;"
+                                                        "java.util.Map"]}]})))
 
 (def java-net-http-classes
   "These classes must be initialized at run time since GraalVM 22.1"
--- a/src/babashka/impl/classes.clj
+++ b/src/babashka/impl/classes.clj
@@ -183,11 +183,9 @@
                             {:methods [{:name "getBundle"
                                         :parameterTypes ["java.lang.String","java.util.Locale",
                                                          "java.lang.ClassLoader"]}]})
-    features/graal? (assoc `org.graalvm.nativeimage.ProcessProperties
-                           {:methods [{:name "exec"
-                                       :parameterTypes ["java.nio.file.Path"
-                                                        "[Ljava.lang.String;"
-                                                        "java.util.Map"]}]})))
+    (resolve 'org.graalvm.nativeimage.ProcessProperties)
+    (assoc `org.graalvm.nativeimage.ProcessProperties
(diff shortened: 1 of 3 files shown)


src/babashka/impl/classes.clj
```diff
@@ -182,7 +182,12 @@
                             `java.util.ResourceBundle
                             {:methods [{:name "getBundle"
                                         :parameterTypes ["java.lang.String","java.util.Locale",
-                                                         "java.lang.ClassLoader"]}]})))
+                                                         "java.lang.ClassLoader"]}]})
+    features/graal? (assoc `org.graalvm.nativeimage.ProcessProperties
+                           {:methods [{:name "exec"
+                                       :parameterTypes ["java.nio.file.Path"
```
borkdude, line 188: I think we can leave out the param types so we can reach both arities

src/babashka/impl/classes.clj
```diff
@@ -182,7 +182,12 @@
                             `java.util.ResourceBundle
                             {:methods [{:name "getBundle"
                                         :parameterTypes ["java.lang.String","java.util.Locale",
-                                                         "java.lang.ClassLoader"]}]})))
+                                                         "java.lang.ClassLoader"]}]})
+    features/graal? (assoc `org.graalvm.nativeimage.ProcessProperties
```
borkdude, line 186: Sorry, I didn't mean a feature flag, but just a check like this:

https://github.com/babashka/babashka/pull/1558/files#diff-10889a6dbab396cc4a07d142a6ddc22e589081f73c25b382b14a5b0b6a3560b1R15

src/babashka/impl/classes.clj
lread: Sure, can update.

src/babashka/impl/classes.clj
lread: Son of a... Ok, will update.

src/babashka/impl/classes.clj
```diff
@@ -183,11 +183,9 @@
                             {:methods [{:name "getBundle"
                                         :parameterTypes ["java.lang.String","java.util.Locale",
                                                          "java.lang.ClassLoader"]}]})
-    features/graal? (assoc `org.graalvm.nativeimage.ProcessProperties
-                           {:methods [{:name "exec"
-                                       :parameterTypes ["java.nio.file.Path"
-                                                        "[Ljava.lang.String;"
-                                                        "java.util.Map"]}]})))
+    (resolve 'org.graalvm.nativeimage.ProcessProperties)
```
borkdude, line 186: Can you move the resolve to the top level? I know this is already the top level, but having resolve at the non-top level is kind of "dangerous" for bb since it will massively bloat the image size. Therefore I'd like to have all `resolve` usage on the top level like the other ones.

lread: Ya sure.

borkdude:
Thanks!

lread:
And thank you for carefully reviewing!

PR #1560, opened 2023-05-13
tests: run babashka.process lib tests on windows

lread:
Now that babashka.process tests are os agnostic we can also run them on Windows.

Retrying #1559

### DIFF
--- a/test-resources/lib_tests/babashka/run_all_libtests.clj
+++ b/test-resources/lib_tests/babashka/run_all_libtests.clj
@@ -93,13 +93,13 @@
   (test-doric-cyclic-dep-problem))
 
 ;;;; babashka.process
-(when-not (windows?)
-  ;; test built-in babashka.process
-  (test-namespaces 'babashka.process-test)
+;; test built-in babashka.process
+(test-namespaces 'babashka.process-test)
 
-  ;; test babashka.process from source
-  (require '[babashka.process] :reload)
-  (test-namespaces 'babashka.process-test))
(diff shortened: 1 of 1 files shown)


lread:
Resubmitted unchanged so we can witness failure on appveyor.

lread:
Ah. Probably `BABASHAKA_TEST_ENV` is not `jvm` when running jvm tests on appveyor.

lread:
Ah no, that's not it, appveyor only runs native libs tests.
Just 2 failures on babashka.processs `process-dir-option-test`.
Probably just a Windows-ism I missed when running tests from babashka project.
I'll retry on my Windows VM and see what I can see.

lread:
Ah. It's a babashka.process windows `program` resolution issue.

Babashka.process `program` resolution always resolves from the current dir.

When running babashka.process tests from babashka we use the CI built `./bb` (as opposed to `bb` on the path when running tests from babashka.process).

When `:dir` is specified it currently plays no part in the program resolution.
It probably should for Windows and cross-platform script compatibility sake.

The failing test is using `:dir` to a subdir and looking for the CI built `../../../bb`.
On Windows we'd like this to resolve to `../../../bb.exe` but since we are resolving from the current dir, and not `:dir` resolution fails.

On macOS and Linux resolution of `program` `../../../bb` actually fails, but failed resolution returns `program` as is, which works fine.

I'll raise an issue on the babashka.process project to consider options.

I'll close this PR. We can re-attempt after addressing in babashka.process.

PR #1574, opened 2023-06-07
Add malli-cli to projects

borkdude:


borkdude:
@piotr-yuxuan There is no need to update libraries.csv, this file is for another purpose. The rest seems good to me!

borkdude:
@piotr-yuxuan There is no need to update libraries.csv, this file is for another purpose. The rest seems good to me!

PR #1578, opened 2023-06-15
Fix README.md prepending the command with bb

tvedeane:
Maybe I'm missing something, but I think the `.clj` file cannot be started without `bb` as the command name.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ File `pst.clj`:
 ```
 
 ``` shell
-$ pst.clj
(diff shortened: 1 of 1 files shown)


borkdude:
True, although you can run it with `./pst.clj` due to the shebang.

tvedeane:
> True, although you can run it with `./pst.clj` due to the shebang.

Right. I just found that out :) (but it also requires that the file is set to be executable).

PR #1673, opened 2024-02-18
Suggest using mise instead of asdf

anderso:
This PR adds note on asdf performance, suggests mise instead, and adds mise installation instructions.

It seems like a shame when so much work has been put into getting bb to start in 20 ms that asdf adds 120-150 ms. And this is not obvious behavior, new bb users will probably just draw the conclusion that bb is very slow to startup instead of suspecting the real culprit, so I think documenting this is warranted.

Sources, the first is from asdf maintainer stating 150 ms for exec, the latter is mise stating 120 ms:

https://stratus3d.com/blog/2022/08/11/asdf-performance/

https://mise.jdx.dev/dev-tools/comparison-to-asdf.html#performance

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,12 +212,24 @@ linux binary.
 
 ### asdf
 
-[asdf](https://github.com/asdf-vm/asdf) is an extendable version manager for linux and macOS.
+[asdf](https://github.com/asdf-vm/asdf) is an extendable version manager for linux and macOS. Note that asdf will add significant startup time to any babashka script, consider using [mise](#mise) instead.
 
 Babashka can be installed using a plugin as follows:
 
     asdf plugin add babashka https://github.com/pitch-io/asdf-babashka
     asdf install babashka latest
+
+### mise
+
+[mise](https://mise.jdx.dev/) is a development environment setup tool for linux and macOS.
+
+Install:
+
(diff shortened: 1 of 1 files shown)


borkdude:
So mise cuts 30 ms from the startup time of asdf? That is better but not that much compared to using vanilla bb?

anderso:
@borkdude no that was bad wording on my part. Mise claims that asdf overhead is 120 ms, and mise adds 4-14 ms but only when changing directory. And each execution should have zero overhead on mise.

borkdude:
OK, merged. I'll ping @pesterhazy here since he's an asdf user and curious to hear his take on this.

anderso:
I was surprised by it when I used asdf, I don't remember the exact numbers I got on bb with asdf but it was immediately noticable in cli usage. I believe executing one simple statement took between 150-200ms.

pesterhazy:
Yeah, it's true that asdf has a hefty performance perf penalty.

I've sometimes added something along those lines to my `.envrc`:

```
export PATH="$(asdf where babashka)/bin:$PATH"
```

which works around the (very) slow shims.

Using `mise` is another good alternative. Mention it in the README is a good idea for a tool that highlights startup time.

Issue #447, opened 2020-05-22
expose flatland.ordered/ordered-map and data-reader


borkdude:
There's maybe a few more data readers to expose:

```
{ordered/set #'flatland.ordered.set/into-ordered-set, ordered/map #'flatland.ordered.map/ordered-map, datascript/Datom #'datascript.db/datom-from-reader, datascript/DB #'datascript.db/db-from-reader, xml/ns #'clojure.data.xml.name/uri-symbol, xml/element #'clojure.data.xml.node/tagged-element}
```

PR #1693, opened 2024-05-13
Add ITransientSet and flatland.ordered-set

### LINKED ISSUE #1692: Feature request: Expose flatland.ordered.set/ordered-set
Follow-up to #447, could we get support for `flatland.ordered.set`?

```
$ rlwrap bb
[babashka] WARNING: task(s) 'run' override built-in command(s). Use :override-builtin true to disable warning.
Babashka v1.3.189 REPL.
Clojure rocks, Bash reaches.

user=> (require '[flatland.ordered.map :as om])
nil
user=> (require '[flatland.ordered.set :as os])
java.lang.Exception: Unable to resolve classname: clojure.lang.ITransientSet [at flatland/ordered/set.clj:4:3]
```

NoahTheDuke:

Closes #1692.

I am not able to run the tests locally (I get NPEs in `babashka.main/fn`), so my apologies if this is broken. I don't know enough about Babashka to solve the tests issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ A preview of the next release can be installed from
 ## Unreleased
 
 - Fix [#1688](https://github.com/babashka/babashka/issues/1688): use-fixtures should add metadata to `*ns*`
+- Fix [#1692](https://github.com/babashka/babashka/issues/1692): Add support for ITransientSet and org.flatland/ordered-set
 
 ## 1.3.190 (2024-04-17)
 
--- a/feature-yaml/babashka/impl/ordered.clj
+++ b/feature-yaml/babashka/impl/ordered.clj
@@ -7,7 +7,7 @@
 (def omap-ns (sci/create-ns 'flatland.ordered.map nil))
 (def oset-ns (sci/create-ns 'flatland.ordered.set nil))
 
-(def ordered-ns
+(def ordered-map-ns
   {'ordered-map (sci/copy-var omap/ordered-map omap-ns)})
 
 (def ordered-set-ns
--- a/feature-yaml/babashka/impl/ordered.clj
+++ b/feature-yaml/babashka/impl/ordered.clj
@@ -1,9 +1,14 @@
 (ns babashka.impl.ordered
   {:no-doc true}
   (:require [flatland.ordered.map :as omap]
+            [flatland.ordered.set :as oset]
             [sci.core :as sci]))
 
 (def omap-ns (sci/create-ns 'flatland.ordered.map nil))
+(def oset-ns (sci/create-ns 'flatland.ordered.set nil))
 
-(def ordered-map-ns
+(def ordered-ns
   {'ordered-map (sci/copy-var omap/ordered-map omap-ns)})
+
+(def ordered-set-ns
+  {'ordered-set (sci/copy-var oset/ordered-set oset-ns)})
--- a/src/babashka/impl/classes.clj
+++ b/src/babashka/impl/classes.clj
@@ -600,6 +600,7 @@
                       clojure.lang.IRef
                       clojure.lang.ISeq
                       clojure.lang.IPersistentVector
+                      clojure.lang.ITransientSet
                       clojure.lang.ITransientVector
                       clojure.lang.Iterate
                       clojure.lang.LispReader$Resolver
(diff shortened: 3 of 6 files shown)


NoahTheDuke:
(Sorry for the force push, it won't happen again, just wanted to fix the temp commit message before you reviewed.)

borkdude:
This PR seems to conflate a few things:

- Exposing stuff like described in the issue
- Changes to the pod tests, not sure what it has to do with the issue
- Exposing ITransientSet, not sure what it has to do with the issue

You may also want to bump this dependency to the newest release including the hashCode fix you made, which requires a new mvn release.

NoahTheDuke:
I can bump the version, sure.

> Exposing ITransientSet, not sure what it has to do with the issue

Babashka can't use `ordered-set` as Babashka doesn't expose `ITransientSet`, which is used in ordered-set. I think that the most minimal change would be to include `ITransientSet` in Babashka and not expose ordered-set, but including it helps everyone.

https://github.com/clj-commons/ordered/blob/574e47018623b9f1b09b0f9b754b02d6633f3437/src/flatland/ordered/set.clj#L122-L143

> Changes to the pod tests, not sure what it has to do with the issue

I copied the changes from your original PR #447 where you changed the pod tests. I can remove those changes if they're not appropriate.

borkdude:
> Babashka can't use ordered-set as Babashka doesn't expose ITransientSet, which is used in ordered-set. I think that the most minimal change would be to include ITransientSet in Babashka and not expose ordered-set, but including it helps everyone.

This isn't true as exposing a built-in dependency doesn't require running the source of that library through the interpreter.
I'll accept it though it's not strictly necessary, as it probably will help someone else in some point in the future.

> I can remove those changes if they're not appropriate.

Not necessary to make any changes on those.

NoahTheDuke:
> This isn't true as exposing a built-in dependency doesn't require running the source of that library through the interpreter.

When I try to use `org.flatland/ordered-set` with babashka right now, it says it can't find `ITransientSet`. Is that not actually the bug here?

borkdude:
The crux is in "right now". When you require it "right now" when adding that dep to bb.edn, it will run the library from source. But after this PR, it won't.

NoahTheDuke:
Oh I see, that makes sense.

borkdude:
Tests are failing with:

```
ERROR in (ordered-test) (test_utils.clj:82)
expected: (= (ordered-set :a 1 :b 2) (bb nil "(flatland.ordered.map/ordered-set :a 1 :b 2)"))
  actual: clojure.lang.ExceptionInfo: ----- Error --------------------------------------------------------------------
Message:  Could not resolve symbol: flatland.ordered.map/ordered-set
Data:     {:type :sci/error, :line 1, :column 1, :file "<expr>", :phase "analysis"}
Location: <expr>:1:1
Phase:    analysis
```

Btw, you can run tests locally as well with `lein do clean, test :only ...`

NoahTheDuke:
I've released org.flatland.ordered 1.15.12, but it looks like Ordered Sets also use `resolve`: https://github.com/clj-commons/ordered/blob/574e47018623b9f1b09b0f9b754b02d6633f3437/src/flatland/ordered/set.clj#L15

Will this be a problem with including them in Babashka?

borkdude:
> Will this be a problem with including them in Babashka?

Yes, why wouldn't it be this time?

borkdude:
Here's the fix I made for the ordered-map: https://github.com/clj-commons/ordered/commit/dd4bfd028556e5717f8d793050597f8e217ec5ea
The fix for ordered-set would be similar

NoahTheDuke:
> Yes, why wouldn't it be this time?

I'm double checking that you're aware before you merge. I can fix that in ordered before updating here.

borkdude:
Yeah, this PR won't get merged with that change as this will blow up the binary size of bb and its compilation time during the builds.

NoahTheDuke:
Let me know what you think: https://github.com/clj-commons/ordered/pull/76

borkdude:
See comments, tl;dr: this looks good, I'll merge tomorrow

borkdude:
Thanks a lot. You can test this with:

``` clojure
bash <(curl https://raw.githubusercontent.com/babashka/babashka/master/install) --dev-build --dir /tmp
```

once the CI build finishes

NoahTheDuke:
It works great, thank you.

Issue #1695, opened 2024-05-23
command line separation with -- broken in several context


borkdude:
Given a script `cmd.clj`:

``` clojure
$ cat /tmp/cmd.clj
(prn *command-line-args*)
```

The following should not have the hyphen probably?

```
$ bb -f /tmp/cmd.clj -- asdf sdf
("asdf" "sdf")
```

```
$ bb /tmp/cmd.clj -- asdf sdf
("--" "asdf" "sdf")
```

```
$ bb -- asdf sdf <<< '*command-line-args*'
nil
```

```
$ bb repl -- asdf sdf <<< '*command-line-args*'
----- Error --------------------------------------------------------------------
Type:     java.lang.Exception
Message:  File does not exist: asdf
```

bobisageek:
I agree (I think) that in all of these cases, the `--` should not be an arg, and `*command-line-args*` should align with the first example.
The `repl` example is sort of interesting, but I guess at the end of the day, the idea of passing args into a REPL isn't really any different.

felixdo:
Hmm, we have -- in use in git, also in kubectl. For those to work I guess they need -- as an arg no?

bobisageek:
Regarding scenario 2 (file name without `-f`), this appears to be the special handling for files in `main/parse-opts`. If the arg is a file, https://github.com/babashka/babashka/blob/5cd2b9d6e60485c16cf849df4667f468f897dfdc/src/babashka/main.clj#L751 gets called, which puts everything after the file name (including `--`) into `:command-line-args`.
The first idea that comes to mind is doing a recur thing sort of like how subcommands are handled, except maybe assoc'ing `:file` onto opts-map, and recurring with `(next options)` and the updated opts-map. I'm not too sure yet if that has unwanted consequences, because `parse-file-opt` is called from two places, so tweaking that could be problematic.

There's another potential scenario, which is a file that doesn't exist (`bb i-dont-exist.clj -- asdf`), and can get set to `:file` if we get to the fallthrough at https://github.com/babashka/babashka/blob/5cd2b9d6e60485c16cf849df4667f468f897dfdc/src/babashka/main.clj#L680-L684
I think this is actually what's happening in scenario 4, because if I'm working this out correctly, https://github.com/babashka/babashka/blob/5cd2b9d6e60485c16cf849df4667f468f897dfdc/src/babashka/main.clj#L614-L618
[...]

PR #1697, opened 2024-05-25
make CLI args available to REPL

bobisageek:
This change is intended to address a couple of issues described in #1695 

### Change 1 - `--repl` should consume one argument (itself) from the arg list
Scenario 4 from the issue:
```
$ bb repl -- asdf sdf <<< '*command-line-args*'
Type:     java.lang.Exception
Message:  File does not exist: asdf
```

**Problem Statement**
`--` does not cause args to stop being processed as bb arguments if it comes after `repl` or `--repl` in the arg list
**Root Cause**
`--repl` consumes two arguments from the arg list
**Fix**
main.clj, line 616: recur with `options` (which has already had `--repl` removed)
**Tests**
Added a test for this combination of args

### Change 2 - make `*command-line-args*` available to REPL
Scenario 3 from the issue:
```
$ bb -- asdf sdf <<< '*command-line-args*'
nil
```

**Problem Statement**
the value of `*command-line-args*` is `nil` at the REPL, even when command line arguments are provided
**Root Cause**
The REPL paths in babashka.main do not bind `*command-line-args*`...
**Fix**
... now they do - main.clj, lines 1024 and 1073
**Tests**
Add tests for the value of `*command-line-args*` in `--repl` mode and with no mode specified (which runs a REPL via a different code branch)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ A preview of the next release can be installed from
 - Fix [#1688](https://github.com/babashka/babashka/issues/1688): use-fixtures should add metadata to `*ns*`
 - Fix [#1692](https://github.com/babashka/babashka/issues/1692): Add support for ITransientSet and org.flatland/ordered-set
 - Bump org.flatland/ordered to `1.15.12`.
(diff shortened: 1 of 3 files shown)


borkdude:
Perfect, thanks!

PR #1699, opened 2024-05-30
Fix #1696: make transcriptor work

### LINKED ISSUE #1696: Make transcriptor work
```
$ bb -Sdeps '{:deps {com.cognitect/transcriptor {:mvn/version "0.1.5"}}}' -e "(require '[cognitect.transcriptor :as xr :refer (check\!)])"
Message:  Unable to resolve classname: clojure.lang.DynamicClassLoader
```

bobisageek:
This potentially closes #1696 

I think this is the first time I've really created "core"-ish stuff, so I'm expecting I might need to change some things, but I can't find anything I've broken with these changes. :smile: 

**Change Descriptions**
- Transcriptor instantiates a `clojure.lang.DynamicClassLoader`, so that class is added to `classes.clj` to make constructors available
- Transcriptor binds `*source-path*`, so that var is added to `clojure.core` in `babashka.impl.clojure.core`. This var isn't used for anything, and isn't bound based on `*file*` or anything, based on the conversation in https://github.com/babashka/sci/issues/922
- Transcriptor calls `clojure.main/with-read-known`, so that macro is added (straight from clojure) to `babashka.impl.clojure.main` and added to the `clojure.main` bb namespace
- Transcriptor calls `clojure.core.server/repl-read`. From digging around, my read was that the impl for this would be the read function defined in `babashka.impl.repl/repl`, so I pulled that function out so I could re-use it from `babashka.impl.server` (sort of following the pattern from other functions in that ns.
- Added some trivial tests for the newly-exposed vars. These tests feel kind of small/limited to me, but the functions are fairly specific.

**Does Transcriptor work?**
Apart from the `*source-path*`/error reporting stuff discussed previously, Transcriptor seems to work as in clojure:

**abc.repl (repl file used in both tests, with an intentional error)**
```clojure
(require '[cognitect.transcriptor :refer [check! on-exit]])

(on-exit #(println "we're done now"))
(+ 1 2)
(check! #{4})
```

**clj**
```clojure repl
$ clj -Sdeps '{:deps {com.cognitect/transcriptor {:mvn/version "0.1.5"}}}'
Clojure 1.11.3
user=> (require '[cognitect.transcriptor :as t])
nil
user=> (t/repl-files ".")
("./abc.repl")
user=> (t/run "abc.repl")
(comment {:transcript "abc.repl", :namespace cognitect.transcriptor.t_1})
; ... transcriptor output...

Execution error - invalid arguments to cognitect.transcriptor.t-1/eval193 at (abc.repl:5).
3 - failed: #{4}
```

**bb**
```clojure repl
$ bb -Sdeps '{:deps {com.cognitect/transcriptor {:mvn/version "0.1.5"}}}'
Babashka v1.3.191-SNAPSHOT REPL.
Clojure rocks, Bash reaches.

user=> (require '[cognitect.transcriptor :as t])
nil
user=> (t/repl-files ".")
("./abc.repl")
user=> (t/run "abc.repl")
(comment {:transcript "abc.repl", :namespace cognitect.transcriptor.t_1})
; ... transcriptor output

clojure.lang.ExceptionInfo: Transcript assertion failed! 3 - failed: #{4}
 [at <repl>:5:1]
```

**Other notes**
- Locally for me, these changes add ~15KB to the executable


bobisageek:
I'll probably be asleep when CI finishes, so I'll check in on the builds in the morning (Eastern US time)

borkdude:
These are some amazing changes, congrats!
Perhaps for full compatibility we can bind `*source-path*` to the correct thing as well then. It doesn't seem like a huge performance penalty.

borkdude:
To make that work, perhaps we do have to move the var to SCI then since `*file*` is bound in `sci/impl/load.cljc` and `*source-path*` should mirror this. I'm not 100% convinced that it's worth it since it seems to be used seldomly. Perhaps we could just add `{'*source-path* sci/file}` to the config as a compromise?

bobisageek:
> Perhaps we could just add `{'*source-path* sci/file}` to the config as a compromise?

That seems like a great middle ground. I was tunnel-visioned on the differences, but this actually gives transcriptor even better parity in bb, because of course binding `*source-path*` now effectively binds `*file*`, which means that the error reports that were reporting "<repl>" now report the path that was passed to `transcriptor/run`.

borkdude:
Thanks again!

PR #1702, opened 2024-06-08
Fix #1700: catch exceptions from resolving symbolic links during `bb.edn` lookup

### LINKED ISSUE #1700: input redirection from shell can cause an error on startup
This is not a high priority to me or anything; just opening this since I stumbled across it.

**version**
1.3.190 (and also master)

**platform**

linux

**problem**

`bb <([some command here])` causes a NoSuchFileException (on my system, at least).

**repro**

```
$ bb <(echo '(println "hi")')
Exception in thread "main" java.nio.file.NoSuchFileException: /dev/fd/63
	at java.base@22/sun.nio.fs.UnixPath.toRealPath(UnixPath.java:886)
	at babashka.fs$real_path.invokeStatic(fs.cljc:80)
	at babashka.fs$real_path.invoke(fs.cljc:76)
	at babashka.fs$real_path.invokeStatic(fs.cljc:78)
	at babashka.main$resolve_symbolic_link.invokeStatic(main.clj:1144)
	at babashka.main$resolve_symbolic_link.invoke(main.clj:1143)
	at babashka.main$main.invokeStatic(main.clj:1171)
	at babashka.main$main.doInvoke(main.clj:1148)
...
```
**expected behavior**

The output of the shell command would be handled as a file that's 'ephemeral', e.g.
```
$ clj -M <(echo '(println "hi")')
hi
```

**notes**

After a bit of digging, this appears to stem from `babashka.main/main`:
https://github.com/babashka/babashka/blob/306842857803f272f0ac900f54710c61e5717b06/src/babashka/main.clj#L1181-L1182

... the redirection in use `<()` creates an ephemeral file (or something along those lines) which I guess doesn't exist by the time bb tries to get the real path. I'm not sure if there's a lot to really do about this without messing with #1661 , but I figure it couldn't hurt to open the discussion. This can generally be worked around by redirecting in a different way.

bobisageek:
**Change Description**
- catch exceptions when resolving symlinks during `babashka.main/main`

**Tests**
- Since the failing scenario is somewhat esoteric, and getting predictable ephemeral files was a bit tricky with JVM testing, the test only runs on native, non-windows builds. The test feels like a pretty faithful repro of #1700 (failed with the same exception prior to the change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ A preview of the next release can be installed from
 - [#1696](https://github.com/babashka/babashka/issues/1696): add `clojure.core/*source-path*` (points to the same sci var as `*file*`) ([@bobisageek](https://github.com/bobisageek))
 - [#1696](https://github.com/babashka/babashka/issues/1696): add `clojure.main/with-read-known` ([@bobisageek](https://github.com/bobisageek))
 - [#1696](https://github.com/babashka/babashka/issues/1696): add `clojure.core.server/repl-read` ([@bobisageek](https://github.com/bobisageek))
+- [#1700](https://github.com/babashka/babashka/issues/1700): catch exceptions from resolving symbolic links during `bb.edn` lookup ([@bobisageek](https://github.com/bobisageek))
 
 ## 1.3.190 (2024-04-17)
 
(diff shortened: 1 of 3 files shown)


borkdude:
Thanks!

PR #1747, opened 2024-10-13
[describe] add commit SHA to describe

lispyclouds:
Not sure if this works


src/babashka/main.clj
```diff
@@ -261,6 +266,7 @@ Use bb run --help to show this help output.
    (format
     (str/trim "
 {:babashka/version   \"%s\"
+ :babashka/build-sha %s
```
borkdude, line 269: Perhaps we can just call this `:git/sha`

lispyclouds:
i suppose this needs a "" around the SHA? this is intended for parsing i think

borkdude:
Yes, bb describe returns EDN

lispyclouds:
well its null on circle
```edn
{:babashka/version   "1.12.195-SNAPSHOT"
 :git/sha            "null"
 :feature/csv        true
 :feature/java-nio   true
 :feature/java-time  true
 :feature/xml        true
 :feature/yaml       true
 :feature/jdbc       false
 :feature/postgresql false
 :feature/sqlite false
 :feature/hsqldb     false
 :feature/oracledb   false
 :feature/httpkit-client true
 :feature/lanterna false
 :feature/core-match true
 :feature/hiccup     true
 :feature/test-check true
 :feature/spec-alpha false
 :feature/selmer true
 :feature/logging true
 :feature/priority-map true}
```

lispyclouds:
checking the rest

lispyclouds:
@borkdude could you check if the cirrus and appveyor works if possible?

borkdude:
@lispyclouds See my comment that you (accidentally) marked as resolved.
Also I think in a PR the commit sha might be under a different env variable, but falling back to the git command (see comment) may take care of that as well.

lispyclouds:
right, got it

borkdude:
We could fall back on a git command:

```
(-> (clojure.java.shell/sh "git" "rev-parse" "HEAD")
          :out
          str/trim)
```

and not actually throw an error if `git` doesn't exist, so you need to wrap this in a try/catch:

```
user=> (clojure.java.shell/sh "gitx" "rev-parse" "HEAD")
Execution error (IOException) at java.lang.ProcessImpl/forkAndExec (ProcessImpl.java:-2).
error=2, No such file or directory
```

lispyclouds:
steps:
- Introduce BABASHKA_SHA env var
- Add to list in native-image.properties
- In CI populate this env var by CI specific ENV vars and fall back to git rev-parse HEAD in the compile script

lispyclouds:
works as expected, cirrus ci however has the SHA for the base branch from where the PR is. maybe we use the git fallback or this is okay as it should work when merged?

lispyclouds:
manually setting it for cirrus to make it consistent

lispyclouds:
should be good @borkdude

borkdude:
Thanks!

PR #1786, opened 2025-01-01
allow subclasses of Throwable to have instance methods called via interop

bobisageek:
Apart from the addition of `Throwable`, the added tests are aimed at covering a couple of the resolutions. So, there's no app code change that's applicable to `Charset` or `InputStream`, but I wanted to cover a few of the superclasses so that if a regression comes up in the future, it might be more evident whether it's an individual superclass that's impacted, or the whole `:public-class` fn.


borkdude (approved):

borkdude:
Thanks!

PR #1799, opened 2025-03-17
docs: readme: badges: reference bb book

### LINKED ISSUE #1798: Consider having bb badges link to explanation of what they mean
## Currently
We [encourage library authors to include a babashka badge in their README when their library is built into, or compatible with, babashka](https://github.com/babashka/babashka?tab=readme-ov-file#badges). 

When a user clicks on a bb badge, they are directed to https://babashka.org.

## Problem
It's great that we click through to the babashka book, but not so great people can't easily discover what the badge means. 

## Proposal
Have the badges link to an explanation of the badge.

## Source
I proposed the idea to @borkdude on Slack, and he welcomed the issue.

### Option 1: Link to existing bb book sections
The babashka.org book has Library -> Project with a description of the compatible badge.
So we can link `babashka | compatible` to https://book.babashka.org/#_projects. 
Or maybe even better add a new deliberate id to same section: https://book.babashka.org/#compatible

The book also has Library -> Built-in namespaces (https://book.babashka.org/#built-in-namespaces).
Create a new deliberate id to this section https://book.babashka.org/#built-in.
Add a description of the built-in badge.

Near each badge description, link back to bb README badge section so folks can learn how to add badges.

Update bb README badge section markdown to new locations.

### Option 2: Link to a new Badges section in bb book
This has the advantage of one link for both badges https://book.babashka.org/#badges

We could also duplicate (or move) instructions on how to create badges to this section from https://github.com/babashka/babashka/blob/master/README.md#badges.

### Option 3: Link to github repo instead
Have badges click through to https://github.com/babashka/babashka/blob/master/README.md#badges

I like this less. I think the babashka book is a better intro to babashka.

And we have existing badges in the wild that already point to the bb book.

## Impact

What about existing [built-in](https://grep.app/search?regexp=true&q=[^\w+]bb+built-in) and [compatible](https://grep.app/search?regexp=true&q=%5B%5E%5Cw+%5Dbb+compatible) badges? 
It would be nice if they were updated.
I'm happy to trickle in PRs to update existing READMEs.

## Proposal
I like option 2. Whatcha think?

## Next Steps
Happy to do PRs on bb book then bb for your review.

lread:
Give only a quick summary of bb badges in the README, and reference the new full description in bb book.

Closes #1798

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,63 +303,16 @@ Go [here](https://book.babashka.org/#built-in-namespaces) to see the full list o
 A list of projects (scripts, libraries, pods and tools) known to work with babashka.
 
 ## Badges
-<!-- note to editor: it seems a blank line must appear before code blocks within <details> -->
 
-[![bb compatible](/logo/badge.svg)](https://babashka.org)
-
-The babashka compatible badge indicates that a [library can be used as babashka dependency](doc/projects.md).
-
+[![bb compatible](/logo/badge.svg)](https://book.babashka.org#badges)
+The babashka compatible badge indicates that a library can be used as babashka dependency.
 If this is the case for your library, we encourage you to proudly display this badge.
 
-<details><summary>Markdown</summary>
-
-```markdown
-[![bb compatible](https://raw.githubusercontent.com/babashka/babashka/master/logo/badge.svg)](https://babashka.org)
-```
-</details>
-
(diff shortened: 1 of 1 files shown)


borkdude:
Yep, thank you!

PR #1835, opened 2025-06-20
Timbre fixes for trove

borkdude:


ptaoussanis:
Thanks for this Michiel, cheers! :-)

PR #1905, opened 2026-01-06
add `alter-meta!` to preserve `babashka.process` function docstrings

### LINKED ISSUE #1727: Missing docstrings in `babashka.process` ns
**version**
babashka v1.4.192

**platform**

MacOS

**problem**

The docstrings of public API functions in `babashka.process` are missing, especially `shell` and `process`, whose arglists `[& args]` provide no clue to to how to use the function in a REPL session.

**repro**

```
$ bb
Babashka v1.4.192 REPL.
Clojure rocks, Bash reaches.


user=> (meta (requiring-resolve 'babashka.process/shell))
{:ns #object[sci.lang.Namespace 0x5019f5b4 "babashka.process"], :name shell, :arglists ([& args]), :doc nil}

```

**expected behavior**
The `:doc` and  custom`:arglists` metadata keys should be present:
https://github.com/babashka/process/blob/689ab390211a854f7427a9976ba526618d60d0cf/src/babashka/process.cljc#L652

I traced this issue to the `babashka.impl.process` namespace which defines wrappers for the functions in the original ns (instead of using `sci/copy-var`), causing their original metadata to be lost:

https://github.com/babashka/babashka/blob/0c6f8b6806a1a8df316d1833cb159e5261410940/src/babashka/impl/process.clj#L11

conao3:
Fixes #1727

add `alter-meta!` to preserve `babashka.process` function docstrings

```
$ java -jar target/babashka-1.12.214-SNAPSHOT-standalone.jar -e '(meta (requiring-resolve (quote babashka.process/shell)))'
{:ns #object[sci.lang.Namespace 0x68312b85 "babashka.process"], :name shell, :arglists ([opts? & args]), :doc "Convenience function around `process` that was originally in `babashka.tasks`.\n  Defaults to inheriting I/O: input is read and output is printed\n  while the process runs. Defaults to throwing on non-zero exit codes. Kills all\n  subprocesses on shutdown. Optional options map can be passed as the\n  first argument, followed by multiple command line arguments. The\n  first command line argument is automatically tokenized. Counter to\n  what the name of this function may suggest, it does not start a\n  new (bash, etc.) shell, it just shells out to a program. As such, it\n  does not support bash syntax like `ls *.clj`.\n\n  Supported options:\n  - `:continue`: if `true`, suppresses throwing on non-zero process exit code.\n  - see `process` for other options\n\n  Examples:\n\n  - `(shell \"ls -la\")` ;; `\"ls -la\"` is tokenized as `[\"ls\" \"-la\"]`\n  - `(shell {:out \"/tmp/log.txt\"} \"git commit -m\" \"WIP\")` ;; `\"git commit -m\"` is tokenized as `[\"git\" \"commit\" \"-m\"]` and `\"WIP\"` is an additional argument\n\n  Also see the `shell` entry in the babashka book [here](https://book.babashka.org/#_shell)."}
```

A more programmatic approach could also be implemented, but this is the most explicit method.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ A preview of the next release can be installed from
 - Bump `deps.clj`
 - Fix #1901: add `java.security.DigestOutputStream`
 - Fix [#1727](https://github.com/babashka/babashka/issues/1727): add `alter-meta!` to preserve `babashka.process` function docstrings
+- Fix [#1754](https://github.com/babashka/babashka/issues/1754): add note for `graalvm/setup-graalvm`
 
 ## 1.12.213 (2025-12-06)
 
--- a/doc/build.md
+++ b/doc/build.md
@@ -87,6 +87,17 @@ Run `script\uberjar.bat` followed by `script\compile.bat`.
 To compile babashka as a static binary for linux, set the `BABASHKA_STATIC`
 environment variable to `true`.
 
+### Musl and zlib
+
+Static compilation requires musl and zlib. Since pre-built musl distributions
+are no longer available on musl.cc, both libraries need to be built from source.
+
+When using [graalvm/setup-graalvm](https://github.com/graalvm/setup-graalvm)
+GitHub Action with `native-image-musl: 'true'`, musl 1.2.4 and zlib 1.2.13 are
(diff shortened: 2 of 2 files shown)


borkdude:
Thanks!

Issue #1972, opened 2026-05-12
Add user-level config for dependency resolution


dominicfreeston:
**Is your feature request related to a problem? Please describe.**
Recently maven central started rate-limiting its usage much more aggressively. I'm not sure if this is in error or by design, but it made the question of how to set up mirrors for babashka come to the fore, and how to do this efficiently across many projects.

**Describe the solution you'd like**
I'd like a user-level setting, similar to what the clojure tools do, where the repositories can be specified and so alternatives can be specified or even defaults overriden.

I see two avenues:
- Use ~/.clojure/deps.edn the same way the clojure tools do
- Create a ~/.babashka/deps.edn convention

As I understand it, babashka uses tools.deps to resolve its dependencies so using the same approach is likely to be relatively straightforward - the main question I see is whether we want to use the same config or have an independent one.

**Describe alternatives you've considered**
The main alternative is to specify the :mvn/repos in each projects - this can be inconvenient which many repos to maintain, but also precludes certain uses cases like switching out the repo on the CI (e.g. so only approved libs can be deployed to production).

borkdude:
I think it should be `~/.babashka/bb.edn` which gets merged with the project `bb.edn`? With respect for XDG conventions.

dominicfreeston:
@borkdude would you rather we merge the whole of the project file into the whole of the user-level bb.edn? Or just :mvn/repos or now? Also, should we support multiple merging strategies? We could do that via metadata or something, if you wanted to go down that road - and if not, do you want project values to just replace the user-level values?

borkdude:
Alternative. We could also pick up on the `:mvn/repos` from the user deps.edn. I'm not sure, need to think about the alternatives more.

dominicfreeston:
I've opened a PR with the following approach:

- just merge :mvn/repos, leave all other keys out (could be added later, but I think for example merging deps could cause unexpected issues)
- merge all the specified repos, giving priority to project-levels ones (i.e. `(merge user-repos project-repos)`)
- respect XDG_CONFIG_HOME with .config default, fall back to ~/.babashka

This addresses the specific use-case raised here without risking introducing other unexpected behaviour, but keeps the possibility of extending the approach later (e.g. user-level tasks?).

(Disclosure: I used an AI-assistant to write this code, hence the slightly strange journey in the commit history, but I have reviewed and refined the code, not just throwing slop your way for you to review 😄)

borkdude:
[...]

PR #1973, opened 2026-05-12
Allow user-level :mvn/repos

dominicfreeston:

#1972

### DIFF
--- a/src/babashka/impl/deps.clj
+++ b/src/babashka/impl/deps.clj
@@ -2,7 +2,7 @@
   (:require [babashka.deps :as bdeps]
             [babashka.fs :as fs]
             [babashka.impl.classpath :as cp]
-            [babashka.impl.common :refer [bb-edn]]
+            [babashka.impl.common :refer [bb-edn debug]]
             [babashka.process :as process]
             [borkdude.deps :as deps]
             [clojure.edn :as edn]
@@ -38,8 +38,9 @@
     (try
       (edn/read-string (slurp f))
       (catch Exception e
-        (binding [*out* *err*]
-          (println (str "[babashka] WARNING: could not read user bb.edn from " f ": " (.getMessage e))))
+        (when @debug
+          (binding [*out* *err*]
+            (println (str "[babashka] WARNING: could not read user bb.edn from " f ": " (.getMessage e)))))
         nil))))
 
 ;;;; merge deps.edn files
--- a/src/babashka/impl/deps.clj
+++ b/src/babashka/impl/deps.clj
@@ -6,7 +6,6 @@
             [babashka.process :as process]
             [borkdude.deps :as deps]
             [clojure.edn :as edn]
-            [clojure.java.io :as io]
             [clojure.string :as str]
             [sci.core :as sci]))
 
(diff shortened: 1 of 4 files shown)


dominicfreeston:
Tests fail but I'm not sure if they're related to my changes? 🤔

borkdude:
I enable some flaky tests on PR branches which are turned off on main, don't worry about those. Just look at the CI runs.

src/babashka/impl/deps.clj
```diff
@@ -2,14 +2,52 @@
...
+(defn- user-bb-edn-path
+  "Returns the path to the user-level bb.edn file, respecting XDG conventions.
+  Checks in order:
+    1. $XDG_CONFIG_HOME/babashka/bb.edn
+    2. ~/.config/babashka/bb.edn (XDG default)
+    3. ~/.babashka/bb.edn (legacy fallback)"
+  []
+  (let [home (System/getProperty "user.home")
+        xdg-config-home (get-xdg-config-home)
+        xdg-path (io/file (or xdg-config-home (io/file home ".config"))
```
borkdude: I'd prefer if we could use babashka/fs + babashka.fs/exists? etc instead of interop

borkdude:
Can you check if https://github.com/clojure/tools.deps.edn is relevant in any way here? I haven't really looked at it before

src/babashka/impl/deps.clj
dominicfreeston: Makes sense, will update 👍

dominicfreeston:
> Can you check if https://github.com/clojure/tools.deps.edn is relevant in any way here? I haven't really looked at it before

@borkdude if we wanted to use the .clojure/deps.edn as the source of truth instead then we could use it. We do similar things in this PR:
- [resolve-file-location](https://github.com/clojure/tools.deps.edn/blob/3bf83a196302e1a53f999fd791a3607aa6aded65/src/main/clojure/clojure/tools/deps/edn.clj#L148) - it does also respect XDG_CONFIG_HOME but it does not default and give priority to .config, which I _believe_ would be the correct thing to do according to the spec. But I don't mind matching their behaviour if you think that's preferable. They also give priority to CLJ_CONFIG, which I'm not sure would make sense here.
- [merge the maps](https://github.com/clojure/tools.deps.edn/blob/3bf83a196302e1a53f999fd791a3607aa6aded65/src/main/clojure/clojure/tools/deps/edn.clj#L228) - we could use this but might be overkill to pull in the lib just for a merge fn? It also merges the entire file which I've avoided for now.

src/babashka/impl/deps.clj
```diff
@@ -2,14 +2,51 @@
...
 (def dns (sci/create-ns 'babashka.deps nil))
 
+;;;; user-level bb.edn
+
+(defn get-xdg-config-home
+  "Returns the value of XDG_CONFIG_HOME, or nil if not set."
+  []
+  (System/getenv "XDG_CONFIG_HOME"))
+
+(defn- user-bb-edn-path
```
eval, line 16: There's [`fs/xdg-config-home`](https://github.com/babashka/fs/blob/a69ea51f7fb1535c49373ecbddfa86958e7fc020/src/babashka/fs.cljc#L1646-L1654) (since circa 2023-2) that would simplify things here.

dominicfreeston: thanks, done

borkdude:
https://github.com/babashka/babashka/issues/1972#issuecomment-4444061174

PR #1992, opened 2026-07-12
Allow reify of Closeable and AutoCloseable, fixes #1988

### LINKED ISSUE #1988: Add `Closeable` and `AutoCloseable` to allowed reify interfaces
It's often useful to reify `java.io.Closeable` or the more modern `java.lang.AutoCloseable` when writing services that you want to be able to stop in a standard way. The standard Clojure macro `with-open` makes use of these interfaces, calling their `.close` method when the block ends.

Adding these interfaces to `impl-java/src/babashka/impl/reify2/interfaces.clj` would allow them to be used by Babashka. I can put together a PR with tests that implements this.

weavejester:
This PR fixes #1988 and adds `java.io.Closeable` and `java.lang.AutoCloseable` to the list of allowed reify interfaces.

- [X] This PR corresponds to an [issue with a clear problem statement](https://github.com/babashka/babashka/blob/master/doc/dev.md#start-with-an-issue-before-writing-code).

- [X] This PR contains a [test](https://github.com/babashka/babashka/blob/master/doc/dev.md#tests) to prevent against future regressions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ A preview of the next release can be installed from
   - `java.nio.channels.Selector`
   - `java.nio.channels.SelectionKey`
 - [#1987](https://github.com/babashka/babashka/issues/1987): Allow the `remove` method to be used on `Iterator`
+- [#1988](https://github.com/babashka/babashka/issues/1988): Allow `java.io.Closeable` and `java.lang.AutoCloseable` to be reified
 
 ## 1.12.218 (2026-04-20)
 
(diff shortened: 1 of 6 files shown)


borkdude:
This looks good, but CI fails because lib needs bumping, I'll do this locally after merge.

weavejester:
Ahh, of course. I forgot I needed to run additional scripts for impl-java.

borkdude:
I was able to commit to your PR so we could verify it via CI.

weavejester:
Great! I've just cleaned and recompiled Babashka from the master branch, and it now runs all the tests for TeensyP without errors or failure.

borkdude:
That's great. You can say to your users that they can get the test build of babashka with:

```
bash <(curl https://raw.githubusercontent.com/babashka/babashka/master/install)  --dev-build --dir /tmp
```

and then they can run your lib! Hopefully a new bb release coming this week.

PR #2041, opened 2026-08-23
[#2040] Allowlist typed NIO buffers for bulk primitive I/O

### LINKED ISSUE #2040: Allowlist typed NIO buffers (IntBuffer / FloatBuffer / …) for bulk primitive I/O
### Problem

Babashka already exposes a useful slice of NIO (`ByteBuffer`, `MappedByteBuffer`, `FileChannel`, `FileChannel$MapMode`, `byte[]`, `int[]`). What is missing are the **typed buffers** and remaining primitive array classes, so SCI cannot call the **bulk** `Buffer.get(primitive[])` overloads.

The fallback is a SCI loop of per-element `.get`. On a ~9MB float32 blob (2,962 embedding rows) that decode took **38 seconds**. The same data as JSON-of-ints via Cheshire (compiled in bb) loaded in **258ms**. A single compiled `FloatBuffer.get(float[])` / `IntBuffer.get(int[])` is a memcpy of that 9MB — on the order of **milliseconds** (estimate; we did not have the allowlist to measure it in bb). That is ~150× vs the JSON workaround and ~1,000–10,000× vs the interpreted per-element path.

### What is already there

With `BABASHKA_FEATURE_JAVA_NIO` (on by default):

- `java.nio.ByteBuffer` (`:all` / all public methods, including `asIntBuffer` / `asFloatBuffer`)
- `MappedByteBuffer`, `DirectByteBuffer`
- `FileChannel` + `MapMode`
- `ByteOrder`
- `[B`, `[I`

`ByteBuffer.asFloatBuffer` can return a `FloatBuffer`, but SCI then has no allowlisted type to invoke `.get(float[])` on. Decode falls back to ~2.3M interpreted interop calls.

### Proposal

Add to the NIO class set (same `:all` / all-public-methods treatment as `ByteBuffer`):

- `java.nio.IntBuffer`
- `java.nio.FloatBuffer`
- `java.nio.LongBuffer`
- `java.nio.DoubleBuffer`
- `java.nio.ShortBuffer`
- primitive array classes `[F`, `[J`, `[D` (`[B` and `[I` already exist)

No new dependency, no new feature flag. These types live in `java.base` and are likely already reachable from `ByteBuffer`; this is mostly SCI + Graal reflection so scripts can call the bulk methods.

### Intended use (one compiled fill)

```clojure
(import '[java.nio ByteBuffer ByteOrder]
        '[java.nio.channels FileChannel FileChannel$MapMode]
        '[java.io FileInputStream])

(defn load-f32le
  "Load a row-major little-endian float32 blob into a float[]."
  [^String path]
  (with-open [in (FileInputStream. path)]
    (let [ch (.getChannel in)
          size (.size ch)
          bb (.order (.map ch FileChannel$MapMode/READ_ONLY 0 size)
                     ByteOrder/LITTLE_ENDIAN)
          dest (float-array (quot size 4))]
      (.get (.asFloatBuffer bb) dest)
      dest)))
```

Same pattern for int32 (`asIntBuffer` + `int-array`). Endianness is already covered (`ByteOrder`).

### Why this is a reasonable LCD for bb

It is general-purpose binary I/O, not an ML library:

- numeric dumps, image headers, simple tensor/embedding blobs
- mmap + bulk copy without leaving bb
- anything scripts already `read-bytes` but cannot unpack quickly

A pod does not substitute well here: returning a 9MB primitive array through the pod protocol re-serializes it (JSON/transit/EDN) — the cost this allowlist avoids.

### AI / embeddings

This is the gap that keeps babashka off a common AI path: **load/save embeddings, vectors, and small tensors** as raw binary (float32 / int8 quantized / int32 scaled). Agent CLIs, RAG indexes, and local embedding caches are exactly “read a few MB of primitives and do a dot product.” Today bb can talk HTTP and JSON all day, but a float32 matrix on disk is either 38s or a JSON workaround.

Enabling bulk buffer I/O does not turn bb into NumPy. It does make the **ingest/export** side of that ecosystem viable in a fast-starting script, which is the niche bb already owns.

Scoring/inner loops can still be slower than JVM Clojure (SCI). The 38s → milliseconds win is **decode**. That is the part that currently makes binary formats unusable.

### Weight

Expected native-image growth: small (tens of KB class, not a new library). Startup: ~0. No extra deps.

Happy to add a regression test of `(.get (.asFloatBuffer (ByteBuffer/wrap bytes)) (float-array n))` filling a known pattern if this is welcome.

slagyr:
Fixes #2040.

SCI could call `ByteBuffer.asFloatBuffer` but not `FloatBuffer.get(float[])` (`Method get on class java.nio.ByteBufferAsFloatBufferB not allowed!`). Same shape as existing `CharBuffer` for the `public-class` `instance?` remap, but only `.get` / `.put` are registered (not `:all`).

Adds:

- `.get` / `.put` on `java.nio.IntBuffer` / `FloatBuffer` / `LongBuffer` / `DoubleBuffer` / `ShortBuffer`
- primitive arrays `[F]`, `[J]`, `[D]`, `[S]` (`[B]` and `[I]` were already present)

## Size

`:all` on these types was +275 KiB. get/put-only was measured at **+132 KiB**.

## 1.13.220 re-benchmark (9MB float32, 2962×768, mmap)

| | time |
|---|---|
| 1.13.219 `.getFloat`+`aset` loop | 41.4 s |
| 1.13.220 same loop | 255 ms |
| 1.13.220 Cheshire JSON-of-ints | 153 ms |
| this PR `FloatBuffer.get(float[])` | **2.0 ms** |

Details: https://github.com/babashka/babashka/issues/2040#issuecomment-5542368714

## Tests

`lein test :only babashka.main-test/byte-buffer-test` — JVM, 8 assertions, 0 failures.

Covers heap `asFloatBuffer` / `asIntBuffer` bulk get, bulk put, and `allocateDirect` (the `DirectFloatBufferRU` path).


slagyr:
Bump.
May I get an estimate of if/when this might get merged and released?

slagyr:
Slimmed to `.get` / `.put` only (the +132 KiB shape), plus `[S]`. Benchmark numbers and the get/put answer are on the issue: https://github.com/babashka/babashka/issues/2040#issuecomment-5542368714

borkdude:
Thanks for the issue and PR.

slagyr:
Awesome!  Thanks @borkdude.

PR #2055, opened 2026-08-27
Support UNIX domain sockets in the socket REPL

licht1stein:
## Summary

`bb socket-repl unix://<path>` listens on a UNIX domain socket instead of a TCP port:

```
$ bb socket-repl unix:///tmp/bb.sock
Babashka socket REPL started at /tmp/bb.sock
$ nc -U /tmp/bb.sock
Babashka v1.13.220-SNAPSHOT
user=> (+ 1 2)
3
```

TCP behavior (`bb socket-repl [addr:]port`) is unchanged; the address kind is picked by the `unix://` prefix.

## Use case

A multi-user server running many REPLs (one or more per user). With TCP, every REPL is a localhost port that any user on the machine can connect to — the REPLs can leak between users. With Unix domain sockets, each socket is a file whose permissions restrict access to its owner, so each user's REPLs stay private to them.

## Details

- A stale socket file left behind by a dead process is detected (connect probe fails with `ECONNREFUSED`) and removed before rebinding, so restarts just work.
- Bind failures print a clean one-line error instead of a stack trace:
  - occupied socket: `Cannot start server at /tmp/bb.sock: socket is already in use by another process`
  - missing parent directory: `Cannot start server at /a/b/c.sock: parent directory /a/b does not exist`
- The socket file is removed on server stop and on JVM exit (shutdown hook).
- Implemented in `babashka.impl.clojure.core.server` on `java.net.UnixDomainSocketAddress` / `ServerSocketChannel` (JDK 16+), adapting the accepted channels to the existing socket-REPL accept loop; no new dependencies.

## Testing

- JVM (`clojure -M:main`) and native image: happy path over `nc -U`, stale-socket rebind, occupied-socket error, missing-parent-dir error, TCP regression (`bb socket-repl 1666`).
- CHANGELOG entry added under Unreleased.

## Question on nREPL

I also implemented the same `unix://<path>` support for `--nrepl-server`, but left it out of this PR until the idea of Unix-socket REPL connections is accepted in principle. If this direction is acceptable, I'm happy to follow up with the nREPL counterpart.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ A preview of the next release can be installed from
 
 ## Unreleased
 
+- `bb socket-repl` can listen on a UNIX domain socket: `bb socket-repl unix://<path>`
 - `babashka.ffi`: a C function that takes a struct as an argument, or returns one, without a pointer in between. Define the fields with a layout, such as `[:struct [[:x :int] [:y :int]]]`. Use a map for the field values. `sizeof` and `alignof` now accept layouts. Struct calls use libffi
 - Link a static libffi for FFI calls that the fixed native call shapes cannot make. Set `BABASHKA_LIBFFI` to another archive or `none`
 - Linux: dynamic binaries now require glibc 2.28. The install script selects the static binary on older systems. See ADR 0010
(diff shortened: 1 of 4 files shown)


borkdude:
It's an interesting idea. I added the socket REPL mostly for compat with clojure and it seems clojure doesn't support unix domain sockets currently. Are there any plans for that?

I did merge nREPL improvements, that now does support unix domain sockets. Basically whole of nrepl.core etc is available in bb now.

